PR #9, opened 2017-04-03
jitter expiration

danielhochman:
Reduce CPU burden of Redis expiration during high traffic

TODO:
- [x] test


src/settings/settings.go
```diff
@@ -9,18 +9,19 @@ type Settings struct {
...
+	UseStatsd                  bool   `envconfig:"USE_STATSD" default:"true"`
+	StatsdHost                 string `envconfig:"STATSD_HOST" default:"localhost"`
+	StatsdPort                 int    `envconfig:"STATSD_PORT" default:"8125"`
+	RuntimePath                string `envconfig:"RUNTIME_ROOT" default:"/srv/runtime_data/current"`
+	RuntimeSubdirectory        string `envconfig:"RUNTIME_SUBDIRECTORY"`
+	LogLevel                   string `envconfig:"LOG_LEVEL" default:"WARN"`
+	RedisSocketType            string `envconfig:"REDIS_SOCKET_TYPE" default:"unix"`
+	RedisUrl                   string `envconfig:"REDIS_URL" default:"/var/run/nutcracker/ratelimit.sock"`
+	RedisPoolSize              int    `envconfig:"REDIS_POOL_SIZE" default:"10"`
+	ExpirationJitterMaxSeconds int64  `envconfig:"EXPIRATION_JITTER_MAX_SECONDS" default:"21600"`
```
junr03: Why is the default 6 hours?

danielhochman: i thought it was a sensible default. you want it to be a large enough window to capture changes in traffic patterns, low enough that it doesn't consume a significant amount of memory. we can have a lower default for open source and override in the env if you'd like.

honestly we should move to add support for LRU internally and do a test without any expiration, but that's for another day.

junr03: No, this is fine. I just wanted to understand.

rodaine:
I would like us to convert a lot of the time-related math to using `time.Duration` and the like, but that can come in a later patch.

:shipit:

PR #10, opened 2017-04-04
gofmt before tests

danielhochman:
also put travis on 1.7.5

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ ifeq ("$(GOPATH)","")
 $(error GOPATH must be set)
 endif
 
+SHELL := /bin/bash
 GOREPO := ${GOPATH}/src/github.com/lyft/ratelimit
 
 .PHONY: bootstrap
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -22,8 +23,12 @@ fix_format:
 	script/docs_fix_format
 	go fmt $(shell glide nv)
 
+.PHONY: check_format
+check_format: docs_format
+	@gofmt -l $(shell glide nv | sed 's/\.\.\.//g') | tee /dev/stderr | read && echo "Files failed gofmt" && exit 1 || true
+
 .PHONY: compile
-compile: docs_format
```
danielhochman: was this needed @junr03?

Makefile
```diff
@@ -22,8 +23,12 @@ fix_format:
...
+compile:
```
junr03: yep I think that is better, no need to format the docs on the compile target.

Makefile
junr03: not needed, that was legacy. I think that putting that in the check_format target is better, and makes more sense.

src/redis/cache.go
```diff
@@ -12,8 +12,11 @@ type TimeSource interface {
 	UnixNow() int64
 }
 
+// Interface for a rand Source for expiration jitter
```
junr03: missing `.`

junr03: and below

src/redis/cache_impl.go
```diff
@@ -189,7 +189,6 @@ func (this *timeSourceImpl) UnixNow() int64 {
 }
 
 // rand for jitter
```
junr03: missing `.`

PR #27, opened 2017-11-11
Add Docker Compose File

mjallday:
- Add README note to include platform build flag.
- Add docker compose file to make it trivial to see how to run.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,43 @@
+
+version: "3"
+services:
+
+  redis:
+    image: redis:alpine
+    expose:
+      - 6379
+    ports:
+      - "6379:6379"
+    networks:
+      - data
+    volumes:
+      - redis-data:/data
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+      - data
+    volumes:
+      - redis-data:/data
+
+  ratelimit:
+    build:
+      context: ./
+      dockerfile: Dockerfile
+    ports:
+      - 8080:8080
```
junr03: You need 6070 for the admin port

docker-compose.yml
```diff
@@ -0,0 +1,43 @@
+
```
junr03: nit: extra line

docker-compose.yml
mjallday: this is resolved.

Issue #23, opened 2017-07-24
make bootstrap failing for 'github.com/Sirupsen/logrus'


amogh-plivo:
When I tried to do a 'make bootstrap', I got this error.
 [ERROR]	Update failed for github.com/Sirupsen/logrus: The Remote does not match the VCS endpoint
[ERROR]	Failed to install: The Remote does not match the VCS endpoint
make: *** [bootstrap] Error 1 

The possible cause if that the logrus package has got renamed to 'github.com/sirupsen/logrus' from 'github.com/Sirupsen/logrus'. This should be fixed in glide files.

junr03:
@amogh-plivo I can't replicate this. Both uppercase S and lowercase s work locally.

junr03:
closing for now

rwlincoln:
This issue may only arise if you are using a case-sensitive file system: https://github.com/golang/dep/issues/433#issuecomment-317146614

PR #34, opened 2018-03-11
Changing github.com/Sirupsen/logrus import path to lower-case.

### LINKED ISSUE #33: Lower-case import path for github.com/Sirupsen/logrus
When adding `ratelimit` as a dependency using `dep` it fails with the error:

> Solving failure: No versions of github.com/lyft/ratelimit met constraints:
	master: Could not introduce github.com/lyft/ratelimit@master due to a case-only variation: it depends on "github.com/Sirupsen/logrus", but "github.com/sirupsen/logrus" was already established as the case variant for that project root by the following other dependers:

The `logrus` README states:

> Everything using logrus will need to use the lower-case: github.com/sirupsen/logrus. Any package that isn't, should be changed.

https://github.com/sirupsen/logrus/blob/master/README.md

As requested in #23, please can the import paths be updated to use the lower-case form.

rwlincoln:
Fixes #33 and #23.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -33,7 +33,7 @@ imports:
   subpackages:
   - pool
   - redis
-- name: github.com/Sirupsen/logrus
+- name: github.com/sirupsen/logrus
   version: ba1b36c82c5e05c4f912a88eab0dcd91a171688f
 - name: github.com/stretchr/testify
   version: f390dcf405f7b83c997eac1b06768bb9f44dec18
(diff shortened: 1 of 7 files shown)


junr03:
@rwlincoln thanks for the patch. Any particular reason for moving the goruntime dep? 

Also the build is failing because the generated mocks need to be updated thanks to the loader.New signature change with the goruntime update.

PR #39, opened 2018-06-04
proto: use the proto defined in data-plane-api

junr03:
Fixes https://github.com/lyft/ratelimit/issues/14

In order to not break any clients using the [old proto](https://github.com/lyft/ratelimit/blob/master/proto/ratelimit/ratelimit.proto), Ratelimit is going to support both handlers: the one for the [data-plane-api service definition](https://github.com/envoyproxy/data-plane-api/blob/466c6ebf0e9e4a75eb3b14e97766b6fb8707b4d3/envoy/service/ratelimit/v2/rls.proto#L10) and the [legacy service definition](https://github.com/lyft/ratelimit/blob/0ded92a2af8261d43096eba4132e45b99a3b8b14/proto/ratelimit/ratelimit.proto#L9). 

The legacy handler will accept a request, translate it into a data-plane-api proto request and use the rest of the codebase as if the new handler had been called. This will incur a perf penalty for clients using the legacy proto.


junr03:
@lyft/network-team @rodaine ptal

junr03:
@mattklein123 given that we don't have a deprecation policy in place for ratelimit I think what we should do is the following.

1. Tag current master as v1.0
2. When we merge this PR tag as v1.1. Announce in envoy-users.
3. After a couple months remove the legacy shim and tag as v1.2. The reason I propose leaving the shim on master for a couple months is in case people are running master instead of pinning to a version.

wdyt?

src/service/ratelimit_legacy.go
```diff
@@ -0,0 +1,143 @@
+package ratelimit
```
mattklein123: Why maintain this code at all? Can't you just convert to YAML and back?

mattklein123:
@junr03 yes that plan sounds fine to me.

src/service/ratelimit_legacy.go
junr03: So the reason I went with this route was to potentially squeeze some perf out of manipulating only golang structs, and not having to deal with marshal/unmarshal oddities which I have had to deal with in envoymanager.

But, the approach you suggest is less code to maintain, and less surface for developer error. I pushed that approach below.

README.md
```diff
@@ -29,6 +30,20 @@ applications. Applications request a rate limit decision based on a domain and a
...
+Envoy's data-plane-api defines a ratelimit service proto [rls.proto](https://github.com/envoyproxy/data-plane-api/blob/master/envoy/service/ratelimit/v2/rls.proto).
+Logically the data-plane-api [rls](https://github.com/envoyproxy/data-plane-api/blob/master/envoy/service/ratelimit/v2/rls.proto)
+is equivalent to the [ratelimit.proto](https://github.com/lyft/ratelimit/blob/0ded92a2af8261d43096eba4132e45b99a3b8b14/proto/ratelimit/ratelimit.proto)
+defined in this repo. However, due
+to the namespace differences and how gRPC routing works it is not possible to transparently route the
+legacy ratelimit (ones based in the [ratelimit.proto](https://github.com/lyft/ratelimit/blob/0ded92a2af8261d43096eba4132e45b99a3b8b14/proto/ratelimit/ratelimit.proto)
+defined in this repo) requests to the data-plane-api
+definitions. Therefore, the ratelimit service will upgrade the requests, process them internally as it would
+process a data-plane-api ratelimit request, and then downgrade the response to send back to the client. This means that,
+for a slight performance hit for clients using the legacy proto, ratelimit is backwards compatible with the legacy proto.
```
rodaine: Are we supporting it long-term, or just for this release? IIRC, the discussions around this were to yank out the old proto support with next release, correct? If so, I'd recommend adding a note to that effect.

PR #41, opened 2018-06-25
redis: add the option to use a separate redis pool for per second limits

junr03:
@lyft/network-team Add the option to use a separate redis pool for per second limits. This allows operators to scale the two clusters independently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,9 @@
 - [Request Fields](#request-fields)
 - [Statistics](#statistics)
 - [Debug Port](#debug-port)
+- [Redis](#redis)
+  - [One Redis Instance](#one-redis-instance)
+  - [Two Redis Instances](#two-redis-instances)
 - [Contact](#contact)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
@@ -58,10 +61,10 @@ to give time to community members running ratelimit off of `master`.
 * Install redis-server.
 * Make sure go is setup correctly and checkout rate limit service into your go path. More information about installing
 go [here](https://golang.org/doc/install).
-* In order to run the integration tests using a local default redis install you will also need these environment variables set:
+* In order to run the integration tests using a local redis server please run two redis-server instances: one on port `6379` and another on port `6380`
   ```bash
-  export REDIS_SOCKET_TYPE=tcp
-  export REDIS_URL=localhost:6379
+  redis-server --port 6379 &
+  redis-server --port 6380 &
   ```
 * To setup for the first time (only done once):
   ```bash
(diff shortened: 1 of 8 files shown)


junr03:
cc @danielhochman @mattklein123 @rodaine

README.md
```diff
@@ -352,6 +355,38 @@ $ curl 0:6070/
 
 You can specify the debug port with the `DEBUG_PORT` environment variable. It defaults to `6070`.
 
+# Redis
+
+Ratelimit uses redis as its caching layer. Ratelimit supports two operation modes:
+
+1. One redis server for all limits.
+1. Two redis instances: one for per second limits and another one for all other limits.
```
danielhochman: nit: capitalize Redis

src/redis/cache_impl.go
```diff
@@ -16,7 +16,12 @@ import (
 )
 
 type rateLimitCacheImpl struct {
-	pool                       Pool
+	pool Pool
+	// Optional Pool for a dedicated cache of per second limits.
+	// If this pool is nil, then the Cache will use the pool for all
+	// limits regardless of unit. If this pool is not nil, then it
+	// is used for limits that have a SECOND unit.
```
danielhochman: if i were to set a 3600 second limit it would end up in the per second pool. is that correct?

src/redis/cache_impl.go
```diff
@@ -45,23 +50,33 @@ func unitToDivider(unit pb.RateLimitResponse_RateLimit_Unit) int64 {
...
 	}
 
-	var cacheKey string = domain + "_"
+	var key = domain + "_"
 	for _, entry := range descriptor.Entries {
-		cacheKey += entry.Key + "_"
-		cacheKey += entry.Value + "_"
+		key += entry.Key + "_"
+		key += entry.Value + "_"
 	}
```
rodaine: How often is this method called? If it's only done initially when the descriptors are loaded, then it's fine, but if it's called in the hot path, this is pretty slow and results in a lot of allocations. Should use a pool of byte.Buffer to build the string if that's the case. The [example](https://golang.org/pkg/sync/#Pool) for sync.Pool is pretty much the same.

src/redis/cache_impl.go
junr03: Do you meant a limit of 3600 with a SECOND unit, or a limit that has an expiration of 3600 seconds?

I think you mean the latter. You are right that the current code would need to be modified to handle cases like that, but the nice thing is that we currently don't support that type of unit semantics. So you *can't* specify a limit with an expiration of 3600 seconds.

src/redis/cache_impl.go
junr03: This is executed in the hot path. I'll fix.

src/redis/cache_impl.go
danielhochman: Ah right.

PR #42, opened 2018-06-25
docs: update contact info

junr03:
@lyft/network-team updating the contact section.


junr03:
@mattklein123 ptal, I dont think this needs its own mailing lists, so I am tagging along the envoy contact info. Don't know if that is kosher given this is not in the envoyproxy org, but I think it is the most relevant contact info we have for it.

PR #43, opened 2018-07-23
fix duplicate mv in scripts/install-protoc

mgjv:
While creating a docker build for ratelimit that would run on my Macbook, I noticed that there is an error in the scripts/install-protoc script when running on a system without sudo (which the golang default docker image is). 

This patch should fix that.

The sed script I used to do this has also corrected a missing newline at the end of the file.

@lyft/network-team


mgjv:
Not sure where I'm supposed to tag you, so trying here as well: @lyft/network-team

PR #46, opened 2018-09-05
docker: upgrade docker-compose setup

junr03:
@lyft/network-team previously the ratelimit docker container used the binary built on the host. This leads to incompatibility if the host OS and the container OS don't match. 

This PR adds an additional container `ratelimit-build` which builds the ratelimit binary and shares it via a shared volume to the `ratelimit` container. I also went ahead and cleaned up the whole docker setup a bit, and added readme docs for it.


junr03:
cc @lemccarthy

PR #51, opened 2018-10-01
Configuration to ignore dotfiles.

BenPope:
This allows ratelimit to run on Kubernetes with configuration from a configmap.

TODO:
lyft/goruntime needs tagging at v0.1.9 (or perhaps later since it was a breaking change), and glide up needs to be run.


BenPope:
Superseded by https://github.com/lyft/ratelimit/pull/52

PR #56, opened 2018-11-15
Add X-RateLimit-* response headers as an opt-in feature

armsnyder:
@lyft/network-team

Around August 2018, Envoy added a new field, `headers` to the `RateLimitResponse` proto message, which can be used to give Envoy custom headers to attach to its response:
https://github.com/envoyproxy/envoy/pull/4015

With this PR the ratelimit service will populate this response field with three headers:

| Header  | Description | How it maps to the descriptor statuses |
| ------------- | ------------- | ------------- |
| `X-RateLimit-Limit`  | The number of requests allowed in the configured unit of time | The `CurrentLimit.RequestsPerUnit` value for the `RateLimitResponse_DescriptorStatus` if there is one. If more than one descriptor from the request has an associated limit, then this field is _NOT_ included in the response. |
| `X-RateLimit-Remaining`  | The number of requests remaining before the subsequent request will be rate limited | The smallest `LimitRemaining` value of the `RateLimitResponse_DescriptorStatus`s returned by the Redis cache. |
| `X-RateLimit-Reset`  | The number of seconds until the limit resets and requests will start to succeed | Calculated similarly to the way the Redis client generates descriptor keys, using modulo arithmetic with the `RateLimit.Unit` and current time. If there are multiple descriptors, the one with the fewest remaining requests and longest time to reset is used. |

The behavior of populating these headers is _disabled by default_, so this change is perfectly backwards-compatible. The behavior must be enabled explicitly using the `RESPONSE_HEADERS_ENABLED` environment variable.

The legacy rate limit API also still functions when `RESPONSE_HEADERS_ENABLED` is `true`, but the headers are dropped from the response during the conversion step.


stevenscg:
@armsnyder I'm new this to project and evaluating it for rate-limiting outside of Envoy (for now). I was planning to build up these standard headers from the raw data in the response anyway, so this feature seems very useful.

Could a standard grpc client receiving the response proxy these headers upstream to the caller?

An example for the project readme could help give the feature some visibility.

stevenscg:
@junr03 @armsnyder I was just building out my own client library that similarly constructs these headers and noticed that the value required for `X-RateLimit-Reset` was not already available via the existing response. I also noticed that Lyft does not expose these headers in their own [APIs](https://developer.lyft.com/docs/rate-limits).

It seems like I see `X-RateLimit-Reset` included in most rate limiting schemes, so I wonder why the project originally decided not to calculate the reset time? (performance? lack of business need?)

Regardless, I am still a big fan of this PR!

src/service/ratelimit.go
```diff
@@ -126,11 +134,85 @@ func (this *service) shouldRateLimitWorker(
...
+			}
+		}
+		if limitCount == 1 {
+			response.Headers = []*core.HeaderValue{
+				rateLimitLimitHeader(limitingDescriptor),
+				rateLimitRemainingHeader(limitingDescriptor),
+				rateLimitResetHeader(limitingDescriptor, now),
+			}
+		} else if limitCount > 1 {
+			// If there is more than one limit, then picking one of them for the "X-RateLimit-Limit"
```
stevenscg: @armsnyder I just ran into this same situation (two or more descriptors that match the request) while building my client library. My test case is contrived and might not represent real world usage, but do we have any sense as to whether most users of this project have multiple descriptors that match on a regular basis?

Like you did here, I have no obvious way to decide which set of limiting conditions to show. Maybe the closest to the limit on a literal or percentage basis - the worst case of them all?

The consuming app doesn't know which rate limiting condition they are hitting when given just the headers and I've never seen a provider that enumerated which rate limit I was hitting.

mattklein123:
Sorry for the embarrassingly long delay in getting this reviewed. Let's get this reviewed and merged. I think we need to fix our CI also which is probably out of date. @junr03 are you able to review this? If not I will take a look.

junr03:
My earlier comments were not resolved. I can take ownership of the branch and you can review, once I merge master to bring in #73. I still think it is for the best (especially since what you had to do for #73), to defer proto compilation to go-control-plane like we have done in other places. I can do that in a subsequent PR to this. 

We can land this together monday, I am ooo today.

mattklein123:
@junr03 sg sorry didn't realize you were still tracking. Thank you!

junr03:
@armsnyder (for obvious permissions reasons) I can't take over this PR. I have opened a new PR https://github.com/lyft/ratelimit/pull/77. PTAL if you would like to review, or let me know if you want to work on it. Closing this version for now.

PR #69, opened 2019-03-15
fix example: rate limit is not an array

yuval-k:


yuval-k:
@lyft/network-team 
travis failed because `../../proto/ratelimit/ratelimit.pb.go:25:11: undefined: proto.ProtoPackageIsVersion3`

Not sure why the travis failed.. as I only changed README.md
I suspect that travis is broken (i.e. will fail on any PR).

README.md
```diff
@@ -298,7 +298,7 @@ descriptors:
     descriptors:
       - key: subkey      
         rate_limit:
-          -  requests_per_unit: 300
+             requests_per_unit: 300
```
mattklein123: I will look into the CI issues but can you change this to 2 space indent which I think is what is in the rest of the file? Also I think the same issue happens above? Thank you.

mattklein123:
If you merge master CI should be fixed. Thank you!

junr03:
taking ownership in https://github.com/lyft/ratelimit/pull/79

PR #76, opened 2019-04-08
deps: update several of ratelimit's dependencies

junr03:
Updating Ratelimit's dependencies:
1.  The biggest change is to move the data-plane-api dependency to go-control-plane. I also removed all the protoc scripts, references, etc. No more protos should be compiled in this repo. The exception is the legacy ratelimit proto. But because it is deprecated it will not be updated anymore. Per the [deprecation schedule](https://github.com/lyft/ratelimit#deprecation-of-legacy-ratelimit-proto) (albeit delayed from original dates) the proto will be removed when the 2.0.0 release is tagged.
2. Move the protoc-gen-validate dependency to the envoyproxy org.
3. Update other deps.


junr03:
Fixes https://github.com/lyft/ratelimit/pull/76

PR #84, opened 2019-05-21
Updating DockerFile and adding sample for ambassador

Abhishek627:
Proto needs to be copied for docker to run.

Also, setting some basic variables. And copying data folder with config to the final image. 
Adding an Entrypoint for the docker to start.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,7 @@ WORKDIR /go/src/github.com/lyft/ratelimit
 
 COPY src src
 COPY script script
+COPY proto proto
 COPY vendor vendor
 COPY glide.yaml glide.yaml
 COPY glide.lock glide.lock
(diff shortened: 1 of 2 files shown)


junr03:
@Abhishek627 sorry for the delay. I was OOO the last few weeks. 

As I see it this PR is solving 3 things:
1. Fixing the proto issue with the Dockerfile. I am inclined to just get rid of the Dockerfile. I have invested way more time in making sure the entires setup with docker-compose is stable and actually gives a useful setup (with redis included)
2. Setting up ENV vars in the Dockerfile. I feel similarly about this to point 1.
3. Adding a config for ambassador. Initially I do not know if we should include ambassador specific stuff in ratelimit, rather than improve ambassador side docs/examples to work with this project. Do you have a preference for putting that here? If we end up putting here I would like to have some documentation around it.

lmk what you think! Thanks for the PR.

Abhishek627:
@junr03 I understood your reasons for 1 and 2.  For 3rd, I saw too many people asking whether rate limit service can be used with ambassador and thought that it might make sense to include this too, since I made it work for me. I think we can leave it out. Instead, I'll write maybe a medium article about it and you can add it to bottom of the README. Would that be okay ?

Abhishek627:
https://medium.com/@abhisheksharma_61389/ambassador-lyfts-rate-limit-service-cdf64b4fa371

junr03:
Thank you so much for writing that @Abhishek627! Awesome, this is going to help a bunch of people.

If you agree with both 1, and 2 do you want to close this PR, or update it to actually deleting the Dockerfile? And do you want to open another PR to link your article? Thanks!

Abhishek627:
@junr03 I'll close this one and open a new PR with deleting Dockerfile and edited Markdown.

Issue #57, opened 2018-11-16
Public docker image


akalini:
Is there a public docker image that we can use? I checked dockerhub.com and there are many images there built by 3rd parties, but I could not find one from Lyft.

stevesloka:
@akalini I ran into the same question. To get started, I have a PR which adds a Dockerfile to let the image be built (See #58).

stevesloka:
Now that #58 is in, could we have an automatic build for this? I'm happy to help set that up, but not sure what backend infrastructure is available. For now I'm going to build my own image, but would be nice to have a shared one to reference.

stevenscg:
@stevesloka This seems to be one of the few envoy-related projects missing from the lyft namespace on docker hub. Like you, I'm building my own as a base, but would certainly use a public "official" one if it existed.

psalaberria002:
Any news on this issue? It would be great to have it publicly available.

junr03:
@psalaberria002 we have not allocated resources to make this happen. However, this should not be too hard to accomplished. If there is someone around that would like to setup the CI plumbing to do so (push image to dockerhub on commit to master) we can work with them to get dockerhub credentials on CI to push to Lyft's org.

psalaberria002:
Travis build, test and deploy on merge to master? Do you have any recommended tool or aproach? If you could share an example Lyft repo where an image is pushed to Dockerhub that would be awesome @junr03 .

sachdeva-vivek:
I am trying to build my docker image locally but getting: 

`
Step 5/13 : COPY vendor vendor
COPY failed: stat /var/lib/docker/tmp/docker-builder548667671/vendor: no such file or directory
`

I tried to comment out the above line but then I later I get:

`Step 9/12 : RUN CGO_ENABLED=0 GOOS=linux go build -o /usr/local/bin/ratelimit -ldflags="-w -s" -v github.com/lyft/ratelimit/src/service_cmd
 ---> Running in 832d6140af6f
src/service_cmd/runner/runner.go:10:2: cannot find package "github.com/lyft/ratelimit/proto/ratelimit" in any of:
	/go/src/github.com/lyft/ratelimit/vendor/github.com/lyft/ratelimit/proto/ratelimit (vendor tree)
	/usr/local/go/src/github.com/lyft/ratelimit/proto/ratelimit (from $GOROOT)
	/go/src/github.com/lyft/ratelimit/proto/ratelimit (from $GOPATH)`

Any ideas?

[...]

PR #85, opened 2019-06-10
Ambassador doc

Abhishek627:
Deleting dockerfile as we discussed and updating the documentation with ambassador use case.


sidmalik123:
Hi @Abhishek627 , I just wanted to know the reason for removing the docker file

Abhishek627:
Hi @sidmalik123 ! Please see the reasoning here : https://github.com/lyft/ratelimit/pull/84

sidmalik123:
@Abhishek627 @junr03 I feel like the docker file is useful for people who want to use the project in kubernetes.

stevesloka:
I'm the one who added the `Dockerfile` initially, I wanted it because we're building an image since one doesn't exist today. Related to #57

Abhishek627:
@sidmalik123 @stevesloka I used the same Dockerfile for my project too. For k8, I think kompose can be used to create kubernetes resource file from the docker-compose file. But I agree, If many people are using the Dockerfile, we should leave that in.

junr03:
@Abhishek627 lets keep this PR only to the ambassador blog post. We can discuss the Dockerfile issue elsewhere.

PR #86, opened 2019-06-11
fix docker file

sidmalik123:
Removed `COPY vendor` because the repo doesn't have it, and added `COPY proto` because we need the proto definitions for the project to build. This change can be tested by running `docker build`


junr03:
Thanks for the update @sidmalik123. Do you mind signing the CLA?

sidmalik123:
done @junr03

PR #89, opened 2019-06-18
Allow clients to set HitsAddend to zero 

adilhafeez:
This code change allows users to set HitsAddend to zero. With this capability in place users can query how much of the ratelimit is remaining by passing in hitsAddend as zero.

Added test cases to validate the new behavior.

Change in API behavior: currently if hitsAddend is missing we'd increment limit by 1. We'd need to modify ratelimit request proto to include `required = true`. Also we would need to mention this change in release notes.


adilhafeez:
ptal #service-mesh

adilhafeez:
ptal @junr03

src/redis/cache_impl.go
```diff
@@ -134,7 +134,7 @@ func (this *rateLimitCacheImpl) DoLimit(
 	}
 
 	// request.HitsAddend could be 0 (default value) if not specified by the caller in the Ratelimit request.
-	hitsAddend := max(1, request.HitsAddend)
+	hitsAddend := max(0, request.HitsAddend)
```
freedomljc: If request.HitsAddend is zero, should we just return instead of continuing the process?

junr03:
I have largely opposed this feature making it into ratelimit for a while. IMO making a "query" request to check the limit defeats the purpose of ratelimit. i.e every request is consuming resources in the system, and thus they should count against the limit. The application should not make decisions based on a probe. Instead it should base decisions based on the ratelimit decision, and if inclined the remaining limit which is sent back in the response proto.

What use cases do you have in mind for supporting this feature?

cc @danielhochman

src/redis/cache_impl.go
danielhochman: Agree, this change would break existing callers who expect the default of 1 behavior.

danielhochman:
Can we get a doc update on this change as well? And as Jose said, a better understanding of use case.

lita:
I also think this might not be the best approach.

Rather than doing this, would it be possible update ratelimit's API to return the remaining value in the response rather than just a boolean when they increment the counter? I believe this is how browsers do it. I am guessing they want to know when they are reaching the limit but still remain stateless.

junr03:
@lita the proto already has that information: https://github.com/envoyproxy/envoy/blob/master/api/envoy/service/ratelimit/v2/rls.proto#L83

src/redis/cache_impl.go
adilhafeez: We want the call to go through to see what the current value is in the redis cache for the key.

This change would make setting hitsAddend as required key in proto (need to set `required = true` using [protoc-gen-validate](https://github.com/envoyproxy/protoc-gen-validate)). Current behavior is that we'd add 1 if hitsAddend is missed in ratelimit request.

danielhochman: That would break backwards compatibility of the proto, which should be treated as the last resort for adding or changing existing functionality.

Issue #45, opened 2018-08-20
Unable to start docker image.


lemccarthy:
I cloned the repo, did make boostrap, make compile, and then built the docker image using the dockerfile in the repo. However, when I attempt to start the image, either with `docker-compose up` or `docker run ratelimiter`, I get the following error:

` standard_init_linux.go:190: exec user process caused "no such file or directory"`

I am on Fedora 26 if that is helpful information.

PR #90, opened 2019-06-19
use dockerfile to build the image

adilhafeez:
It looks like Dockerfile is not being used at all during the build and in the docker-compose command. This change updates docker-compose to use Dockerfile to build ratelimit binary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ cover.out
 bin/
 .idea/
 vendor
+.glide
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ WORKDIR /go/src/github.com/lyft/ratelimit
 COPY src src
 COPY script script
 COPY vendor vendor
+COPY proto proto
 COPY glide.yaml glide.yaml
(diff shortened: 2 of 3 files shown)


junr03:
We used to have problems #45 with the docker setup, which is why I created the current docker-compose situation with a two step build/run system in #46. I would rather not revert to a more complicated build system given that the protos are pre-compiled into the repo tree. Is there a reason why you want to see this change go in?

cc @Abhishek627 and @sidmalik123 who have been chatting about the Dockerfile and the need for it. IMO we can have a Dockerfile around. However, given that the protos are precompiled, and all we need is a functioning golang toolchain to build the ratelimit binary, I see maintaining that Dockerfile as an undue burden.

adilhafeez:
I've seen examples of using Dockerfile used to build binary. And I think its somewhat standard to have Dockerfile to produce binary. docker-compose is more like setting up an integration environment that someone could use to run full test suite. And in docker-compose if we setup a separate build then we'd end up duplicating the build instructions that we specified in Dockerfie.

PR #95, opened 2019-08-16
Add Redis database number to settings

anpryl:
@lyft/network-team add ability to pass Redis database number to ratelimit

### DIFF
--- a/README.md
+++ b/README.md
@@ -380,6 +380,8 @@ Ratelimit uses Redis as its caching layer. Ratelimit supports two operation mode
 1. One Redis server for all limits.
 1. Two Redis instances: one for per second limits and another one for all other limits.
 
+1. `REDIS_DB` and `REDIS_PERSECOND_DB` could be used to configure redis database number
+
 ## One Redis Instance
 
 To configure one Redis instance use the following environment variables:
@@ -387,6 +389,7 @@ To configure one Redis instance use the following environment variables:
 1. `REDIS_SOCKET_TYPE`
 1. `REDIS_URL`
 1. `REDIS_POOL_SIZE`
+1. `REDIS_DB`
 
 This setup will use the same Redis server for all limits.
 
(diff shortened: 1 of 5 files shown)


mattklein123:
@junr03 PTAL

junr03:
@anpryl what is the use case for this setting?

anpryl:
@junr03 we have single redis instance for multiple applications. Each application uses own database by number.

junr03:
Thanks for the explanation @anpryl. Three high level comments:
1. We need to add documentation for this feature.
2. We need to add tests for this feature.
3. I would like to work to make this feature use a new NewPool method like @repl-david-winiarski did in #96. It would be ideal if we don't change the "legacy" plain method. This way we have a precedent for acquiring different types of pools from the driver.

anpryl:
Working on it

anpryl:
@junr03 	

> I would like to work to make this feature use a new NewPool method like @repl-david-winiarski did in #96. It would be ideal if we don't change the "legacy" plain method. This way we have a precedent for acquiring different types of pools from the driver.

I'm not sure that I understand you correctly.
Do you propose to add new function like `NewPoolImplWithDatabaseNumber` instead of extending an existing one?
I think that the database number is very generic connection parameter. Both TLS and non-TLS functions could accept it. In other case we need two version of non-TLS `NewPool` and two versions for TLS `NewPool`.
Could you please explain why do we need separate function to accept database number?

anpryl:
@junr03 I reworked `NewPoolImpl` to accept a list of functions to configure redis client. Also, I add `WithDatabase` to configure the database number.
Also, it will be easier to extend the connection configuration (for example: TLS).

Working on tests.

junr03:
@henryyyang for a second pair of eyes.

anpryl:
bump to remove stale tag

HenryYYang:
I just realized that this is for ratelimit service which use a local copy of redis.  I don't have a lot of context on this.  That said, multiple database on a single redis is generally a bad thing with little benefit see this discussion.  https://groups.google.com/forum/#!topic/redis-db/vS5wX8X4Cjg/discussion
A better alternative would be to change the key to create logical partition within a single database, or create multiple redis instances.

anpryl:
@HenryYYang database numbers is stable redis feature. I believe multiple teams already using it. Why not add support of stable feature?

intelliping:
Agree with @anpryl, can we please support Redis db selection? It's a widely used feature in the industry.

Same request has been asked by others: https://github.com/envoyproxy/ratelimit/issues/402

PR #109, opened 2019-12-09
fix: support auth without tls

tangxinfa:
Redis's auth is not related with SSL, a non-ssl connection can have auth too.


tangxinfa:
added a integration test for auth without tls.

tangxinfa:
Need reopen this pull request

PR #111, opened 2019-12-19
Add local cache to store whether it is over the limit

freedomljc:
According to https://docs.google.com/document/d/1rucr2O-zfxiYf3fAZYsap-CCtVnwqKbj6nD-ebCZda8/edit?usp=sharing, add local cache to store whether it is over the the limit, and avoid reading redis cache for the already over-the-limit cache_keys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,7 +376,7 @@ You can specify the debug port with the `DEBUG_PORT` environment variable. It de
 
 # Local Cache
 
-Ratelimit optimally uses [freecache](https://github.com/coocood/freecache) as its local caching layer, which stores the over-the-limit cache keys, and avoid reading 
+Ratelimit optionally uses [freecache](https://github.com/coocood/freecache) as its local caching layer, which stores the over-the-limit cache keys, and thus avoids reading the 
 redis cache again for the already over-the-limit keys. The local cache size can be configured via `LocalCacheSizeInBytes` in the [settings](https://github.com/lyft/ratelimit/blob/master/src/settings/settings.go).
 If `LocalCacheSizeInBytes` is 0, local cache is disabled.
 
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@
 - [Request Fields](#request-fields)
 - [Statistics](#statistics)
 - [Debug Port](#debug-port)
+- [Local Cache](#local-cache)
 - [Redis](#redis)
   - [One Redis Instance](#one-redis-instance)
   - [Two Redis Instances](#two-redis-instances)
@@ -374,6 +375,7 @@ $ curl 0:6070/
 You can specify the debug port with the `DEBUG_PORT` environment variable. It defaults to `6070`.
 
 # Local Cache
+
 Ratelimit optimally uses [freecache](https://github.com/coocood/freecache) as its local caching layer, which stores the over-the-limit cache keys, and avoid reading 
 redis cache again for the already over-the-limit keys. The local cache size can be configured via `LocalCacheSizeInBytes` in the [settings](https://github.com/lyft/ratelimit/blob/master/src/settings/settings.go).
 If `LocalCacheSizeInBytes` is 0, local cache is disabled.
--- a/README.md
+++ b/README.md
@@ -373,6 +373,11 @@ $ curl 0:6070/
 
 You can specify the debug port with the `DEBUG_PORT` environment variable. It defaults to `6070`.
 
+# Local Cache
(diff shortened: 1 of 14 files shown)


glide.lock
```diff
@@ -13,7 +17,7 @@ imports:
 - name: github.com/fsnotify/fsnotify
   version: 629574ca2a5df945712d3079857300b5e4da0236
 - name: github.com/gogo/protobuf
-  version: ba06b47c162d49f2af050fb4c75bcbc86a159d5c
```
HenryYYang: are all the version upgrade for the libraries needed for this change?

freedomljc: Those version changes in this file are auto-added when adding/updating the lib in glide.yaml.

src/settings/settings.go
```diff
@@ -31,6 +31,7 @@ type Settings struct {
 	RedisPerSecondAuth         string `envconfig:"REDIS_PERSECOND_AUTH" default:""`
 	RedisPerSecondTls          bool   `envconfig:"REDIS_PERSECOND_TLS" default:"false"`
 	ExpirationJitterMaxSeconds int64  `envconfig:"EXPIRATION_JITTER_MAX_SECONDS" default:"300"`
+	LocalCacheSize             int    `envconfig:"LOCAL_CACHE_SIZE" default:"0"`
```
mattklein123: What unit is this? Entries? Bytes? More comments and/or make the name more clear?

src/redis/cache_impl.go
```diff
@@ -222,6 +249,12 @@ func (this *rateLimitCacheImpl) DoLimit(
 				// in the near limit range.
 				limits[i].Stats.NearLimit.Add(uint64(overLimitThreshold - max(nearLimitThreshold, limitBeforeIncrease)))
 			}
+			if this.localCache != nil {
```
mattklein123: One thing to point out here is for longer limits, this might effectively double the time that something is rate limited. Meaning, let's say you have an hour limit. If you get over limit at the end of the hour and are about to reset, you would have this entry in the cache for the entire duration of the limit, right?

One option would be to set the cache TTL to only the number of seconds remaining in the limit, since you know the current time and the divider. The other option for now would be to clearly document this case with comments and in the end user docs.

freedomljc: I just had a similar conversation with Henry yesterday. I think it doesn't matter whether setting to be the remaining duration or the entire duration. Based on current design, both the redis cache key and local_cache key are similar to `actualKey_duration`. For example, if we have an hour limit on all mongo connections, the cache key would be similar to `mongo_1h`, `mongo_2h`.  let's say, in the hour 1 (0h0m - 0h59m), we started to get ratelimited in the 50th minutes, and the ttl of local_cache will be set as 1 hour(0h50m-1h49m), the cache key is `mongo_1h`. In the time of 1h1m, since either the local cache key or redis cache key becomes different (mongo_2h), it won't get ratelimited.
Here is how cache key gets generated: https://github.com/lyft/ratelimit/blob/8d923502fb14fa2b2d31c945676dcea503f11e0b/src/redis/cache_impl.go#L74-L82

Since it requires reading the redis cache to calculate the remaining duration, it should be better to set the entire duration.

I'll add the comment to make it clear.

HenryYYang: In this simpler approach, we store 2x the number of overlimited keys.  The average case is that we have 100s of overlimited rules, but in the worst case the memory bounded LRU cache will evict the old key we no longer need.

mattklein123: I see OK, that makes sense that the cache key changes. Thanks for the explanation. Yeah mor comments would be good as well as the end-user docs.

src/settings/settings.go
freedomljc: Just changed the name, to make it clear.

src/redis/cache_impl.go
freedomljc: just add comments about it.

freedomljc:
@mattklein123 , can you take a look again? thanks.

README.md
```diff
@@ -373,6 +374,12 @@ $ curl 0:6070/
 
 You can specify the debug port with the `DEBUG_PORT` environment variable. It defaults to `6070`.
 
+# Local Cache
+
+Ratelimit optimally uses [freecache](https://github.com/coocood/freecache) as its local caching layer, which stores the over-the-limit cache keys, and avoid reading
```
mattklein123:
```suggestion
Ratelimit optionally uses [freecache](https://github.com/coocood/freecache) as its local caching layer, which stores the over-the-limit cache keys, and thus avoids reading the 
```

PR #114, opened 2019-12-30
add local cache stats

freedomljc:
To further check how local_cache performs, add stats for local_cache.  And the local_cache stats will be available in `{debug_port}/stats`.

### DIFF
--- a/src/redis/local_cache_stats.go
+++ b/src/redis/local_cache_stats.go
@@ -1,0 +1,43 @@
+package redis
+
+import (
+	"github.com/coocood/freecache"
+	stats "github.com/lyft/gostats"
+)
+
+type localCacheStats struct {
+	cache             *freecache.Cache
+	evacuateCount     stats.Gauge
+	expiredCount      stats.Gauge
+	entryCount        stats.Gauge
+	averageAccessTime stats.Gauge
+	hitCount          stats.Gauge
+	missCount         stats.Gauge
+	lookupCount       stats.Gauge
+	overwriteCount    stats.Gauge
+}
+
+func NewLocalCacheStats(localCache *freecache.Cache, scope stats.Scope) stats.StatGenerator {
+	return localCacheStats{
+		cache:             localCache,
+		evacuateCount:     scope.NewGauge("evacuateCount"),
+		expiredCount:      scope.NewGauge("expiredCount"),
+		entryCount:        scope.NewGauge("entryCount"),
+		averageAccessTime: scope.NewGauge("averageAccessTime"),
+		hitCount:          scope.NewGauge("hitCount"),
+		missCount:         scope.NewGauge("missCount"),
+		lookupCount:       scope.NewGauge("lookupCount"),
+		overwriteCount:    scope.NewGauge("overwriteCount"),
+	}
+}
+
+func (stats localCacheStats) GenerateStats() {
+	stats.evacuateCount.Set(uint64(stats.cache.EvacuateCount()))
+	stats.expiredCount.Set(uint64(stats.cache.ExpiredCount()))
+	stats.entryCount.Set(uint64(stats.cache.EntryCount()))
+	stats.averageAccessTime.Set(uint64(stats.cache.AverageAccessTime()))
+	stats.hitCount.Set(uint64(stats.cache.HitCount()))
+	stats.missCount.Set(uint64(stats.cache.MissCount()))
+	stats.lookupCount.Set(uint64(stats.cache.LookupCount()))
+	stats.overwriteCount.Set(uint64(stats.cache.OverwriteCount()))
+}
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -141,6 +141,14 @@ func testBasicBaseConfig(grpcPort, perSecond string, local_cache_size string) fu
 			response)
 		assert.NoError(err)
 
+		// Manually flush the cache for local_cache stats
+		runner.GetStatsStore().Flush()
+		localCacheHitCounter := runner.GetStatsStore().NewGauge("ratelimit.localcache.hitCount")
+		assert.Equal(0, int(localCacheHitCounter.Value()))
+
+		localCacheMissCounter := runner.GetStatsStore().NewGauge("ratelimit.localcache.missCount")
+		assert.Equal(0, int(localCacheMissCounter.Value()))
+
 		response, err = c.ShouldRateLimit(
 			context.Background(),
 			common.NewRateLimitRequest("basic", [][][2]string{{{getCacheKey("key1", enable_local_cache), "foo"}}}, 1))
@@ -155,6 +163,18 @@ func testBasicBaseConfig(grpcPort, perSecond string, local_cache_size string) fu
 		// store.NewCounter returns the existing counter.
 		key1HitCounter := runner.GetStatsStore().NewCounter(fmt.Sprintf("ratelimit.service.rate_limit.basic.%s.total_hits", getCacheKey("key1", enable_local_cache)))
 		assert.Equal(1, int(key1HitCounter.Value()))
+
+		// Manually flush the cache for local_cache stats
+		runner.GetStatsStore().Flush()
+		localCacheHitCounter = runner.GetStatsStore().NewGauge("ratelimit.localcache.hitCount")
+		assert.Equal(0, int(localCacheHitCounter.Value()))
+
+		localCacheMissCounter = runner.GetStatsStore().NewGauge("ratelimit.localcache.missCount")
+		if enable_local_cache {
+			assert.Equal(1, int(localCacheMissCounter.Value()))
+		} else {
+			assert.Equal(0, int(localCacheMissCounter.Value()))
+		}
 
 		// Now come up with a random key, and go over limit for a minute limit which should always work.
 		r := rand.New(rand.NewSource(time.Now().UnixNano()))
@@ -193,6 +213,26 @@ func testBasicBaseConfig(grpcPort, perSecond string, local_cache_size string) fu
 				assert.Equal(i-20, int(key2LocalCacheOverLimitCounter.Value()))
 			} else {
 				assert.Equal(0, int(key2LocalCacheOverLimitCounter.Value()))
+			}
+
+			// Manually flush the cache for local_cache stats
+			runner.GetStatsStore().Flush()
+			localCacheHitCounter = runner.GetStatsStore().NewGauge("ratelimit.localcache.hitCount")
+			if enable_local_cache && i >= 20 {
+				assert.Equal(i-20, int(localCacheHitCounter.Value()))
+			} else {
+				assert.Equal(0, int(localCacheHitCounter.Value()))
+			}
+
+			localCacheMissCounter = runner.GetStatsStore().NewGauge("ratelimit.localcache.missCount")
+			if enable_local_cache {
+				if i < 20 {
+					assert.Equal(i+2, int(localCacheMissCounter.Value()))
+				} else {
+					assert.Equal(22, int(localCacheMissCounter.Value()))
+				}
+			} else {
+				assert.Equal(0, int(localCacheMissCounter.Value()))
 			}
 		}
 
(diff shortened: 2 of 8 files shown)


freedomljc:
Manually tested it, the local_cache stats can be generated:
```
ratelimit.localcache.hitCount: 8
ratelimit.localcache.averageAccessTime: 1577747704
ratelimit.localcache.entryCount: 2
ratelimit.localcache.evacuateCount: 0
ratelimit.localcache.expiredCount: 0
ratelimit.localcache.lookupCount: 56
ratelimit.localcache.missCount: 48
ratelimit.localcache.overwriteCount: 0
```

test/integration/integration_test.go
```diff
@@ -89,7 +89,7 @@ func testBasicBaseConfig(grpcPort, perSecond string, local_cache_size string) fu
 		os.Setenv("RUNTIME_SUBDIRECTORY", "ratelimit")
 		os.Setenv("REDIS_PERSECOND_SOCKET_TYPE", "tcp")
 		os.Setenv("REDIS_SOCKET_TYPE", "tcp")
-		os.Setenv("LOCAL_CACHE_SIZE", local_cache_size)
+		os.Setenv("LOCAL_CACHE_SIZE_IN_BYTES", local_cache_size)
```
freedomljc: In #111 (comment), we updated the env variable name from LOCAL_CACHE_SIZE to LOCAL_CACHE_SIZE_IN_BYTES, but missed updating the env variable name in test case. (It is a duplicate change for https://github.com/lyft/ratelimit/pull/112)

mattklein123: Why don't the integration tests run in CI? I don't remember the history here. Can you fix this? cc @junr03

src/redis/local_cache_stats.go
```diff
@@ -0,0 +1,43 @@
+package redis
```
mattklein123: This file needs actually tests like we do for other stats. Please add coverage (integration or otherwise) that actually make use of the stats and verify them.

test/integration/integration_test.go
freedomljc: The integration tests run in the CI. In current framework of integration test, client sends the requests, and then verifies the response.  When adding the integration test for local_cache, we expect the responses are same as before, no matter whether local_cache is enabled or not. if there is a typo for env variable, it means the local_cache is disabled. That's why it is not easy to catch the typo. But there is one trick we can do to verify whether local_cache is used in the integration test: add time.sleep into it,, check {debug_port}/stats to see local_cache related stats.

src/redis/local_cache_stats.go
freedomljc: Just embedded the test into the existing test case `TestOverLimitWithLocalCache`,  and we can better compare new-added local_cache stats with the existing stats.

freedomljc:
@mattklein123 , when you get the chance, can you take a look again? thanks.

test/integration/integration_test.go
junr03: it all runs in the same travis target: https://travis-ci.org/lyft/ratelimit/builds/631398436#L480. It could be separated into a unit test and an integration test target.

freedomljc: @junr03 , but if I understood correctly, separating it into another target won't help to catch this typo. By mistyping `LOCAL_CACHE_SIZE_IN_BYTES ` as `LOCAL_CACHE_SIZE `, it means the local_cache is disabled. Since we expect the responses are same no matter whether local_cache is enabled or not, the tests on new-added target still pass.

junr03: Yep, I wasn't talking about the typo problem. I was talking about making it easier to see that there are CI targets for both unit tests and integration tests.

test/redis/cache_impl_test.go
```diff
@@ -189,6 +227,57 @@ func TestOverLimitWithLocalCache(t *testing.T) {
...
+	ok, err := regexp.Match(pattern, []byte(sink.record))
+	if err != nil {
+		t.Fatal("Error: ", err)
+	}
+	if !ok {
+		t.Errorf("Expected: '%s' Got: '%s'", pattern, sink.record)
+	}
+
+	// Check the correctness of hitCount, missCount, lookupCount, expiredCount and entryCount
+	submatch := regexp.MustCompile(`localcache.hitCount:(\d+)\|g\n`).FindStringSubmatch(sink.record)
```
mattklein123: There is a huge amount of duplication between each test on how you deal with the regex, etc. Can you unify that into a utility? Also, at a higher layer, testing the stats in this way seems suboptimal. Can't you just store the stats as integers vs. strings and then look at them directly?

freedomljc: Just moved it to the utility method, and also simplified it.

test/integration/integration_test.go
freedomljc: I'm hesitating to make that change. Generally, the developer should be able to setup the unit test and integration test env, and figure out which part goes wrong if the test fails.

freedomljc:
> Thanks this LGTM. Can you add 1 negative integration test which confirms there is no local cache when the environment variable is not / the default? This can be done by just taking one of the existing tests and verifying all of the local cache stats are 0?

Currently the integration tests are more about client sends the ratelimit requests to sever, and verify the ratelimit response.  But there isn't a good way to access the local_cache stats from the client side. To verify the local_cache stats, we'll have to parse the response from the {debug_port}/stats. pls let me know if you prefer to do in this way?

mattklein123:
> pls let me know if you prefer to do in this way(parse the response, check whether the new-added local_cache stats are available or not) ?

The way this is typically handled is to allow grabbing the stats directly from the running server when running for an integration test. Meaning, just reach directly into the server, grab the stats store, and inspect the stats. I just refreshed my memory of this code and it looks like this is not easily possible, though it would make for much better tests in general, if you want to work on it. 

Connecting to the stats endpoint and parsing the output is also a way of doing it though it seems more hacky than necessary.

I think it would be fine to do this as a follow up, but having looked at this change again, I don't understand how the integration testing support you added for the change originally actually tests anything if the tests all pass with the wrong env variable? Shouldn't some integration tests fail if the variable is not set correctly? Or do you not have an integration test with the local cache?

freedomljc:
> I think it would be fine to do this as a follow up, but having looked at this change again, I don't understand how the integration testing support you added for the change originally actually tests anything if the tests all pass with the wrong env variable? Shouldn't some integration tests fail if the variable is not set correctly? Or do you not have an integration test with the local cache?

The integration test with the local cache works, I got the above stats(https://github.com/lyft/ratelimit/pull/114#issuecomment-569825518) by temporarily changing the integration test: adding `time.sleep`, and checking the `{debug_port}/stats`.

Regarding the reasons why the integration test didn't fail with the wrong env variable before: 

- If there is the wrong env variable `LOCAL_CACHE_SIZE`, the setting(https://github.com/lyft/ratelimit/blob/54e6c7cdb9ff312ade1f66b8986c4f719b33fa6d/src/settings/settings.go#L34) will take the default value 0 for `LOCAL_CACHE_SIZE_IN_BYTES `. If the size is 0, local_cache will not be enabled.
- In the integration test, we expected the responses are same as before no matter whether the local_cache is enabled or not.


I'll continue to improve on it when I gets more bandwidth. May ask for your advise later.

mattklein123:
Sorry I'm still confused. In https://github.com/lyft/ratelimit/blob/master/test/integration/integration_test.go you have a bunch of test logic around the local cache. If the test thought the local cache was enabled, but it was not, what are those tests testing? Please remove all of that logic if it doesn't actually test anything.

I think having good integration tests for this is likely as easy as exposing the stats scope outside of the runner: https://github.com/lyft/ratelimit/blob/87ed58e3a8c65548e081d4be7097959d873d28ad/src/service_cmd/runner/runner.go#L22

Basically just make the runner into an object with methods, and store the root stat scope in a way that you can access it from tests and read the stat data. Then your integration tests can actually look at stats to verify the right interactions if there is no other way to verify them.

freedomljc:
Sorry for the confusion. I initially added the integration test(https://github.com/lyft/ratelimit/pull/111/commits/e6cf1ecac13e2461a31ea5bc80d52867dddfaf68),  to validate whether it would change the behavior(in our case, the responses should be same as before). But please let me know if you prefer to remove the added tests.

Thanks for your suggestion, it works (https://github.com/lyft/ratelimit/pull/115/files). Since it changes the interface, prefer to separate it into a different PR.  Will let you know when it is ready for review.

mattklein123:
> Sorry for the confusion. I initially added the integration test(e6cf1ec), to validate whether it would change the behavior if local_cache is enabled.(in our case, the responses should be same as before). But please let me know if you prefer to remove the added tests.

My suggestion is to:
1) First review/merge https://github.com/lyft/ratelimit/pull/115. Let me know when that is no longer WIP.
2) Remove all of the local cache specific logic from the integration tests internals.
3) Instead test the local cache in the integration tests using the stats you can easily get from 115.

freedomljc:
@mattklein123 , I just made two changes in integration test:
- Compare the stats `over_limit_with_local_cache` and `over_limit`, to further verify whether local_cache logic works as expected. 
- Undo the changes on TestBasicConfigLegacy, considering we are not able to validate the stats if the port is not changed(https://github.com/lyft/ratelimit/issues/118).

Regarding your suggestions:
> Remove all of the local cache specific logic from the integration tests internals.

After undoing the changes on TestBasicConfigLegacy, most changes in integration test have been reverted (check the integration_test.go in [diff](https://github.com/lyft/ratelimit/compare/8d923502fb14fa2b2d31c945676dcea503f11e0b...09d62da657f1ccb98125737adf6c0d1df1b2d099)), except for the logic about checking the over_limit and over_limit_with_local_cache when local_cache is enabled.


> Instead test the local cache in the integration tests using the stats you can easily get from 115.

There exists some inconvenience to validate the new-added [local stats](https://github.com/lyft/ratelimit/blob/09d62da657f1ccb98125737adf6c0d1df1b2d099/src/redis/local_cache_stats.go)  in integration test. We'll have to add time.sleep(10s) before validating that stats, considering current statsstore flushes the stats per [10s](https://github.com/lyft/gostats/blob/118d7f89597e39f7e3c9e2b7539bb86960f6e19d/stats.go#L192). I'm inclined not to add sleep(10s) here. But there also exists other options like passing the customized statsstore as the parameter of run() method in runner, or be able to update the statsstore after the construction of runner. But i'm not sure whether it is worthwhile to make those changes considering the parameter statsstore is not needed in most cases.

mattklein123:
> There exists some inconvenience to validate the new-added local stats in integration test. We'll have to add time.sleep(10s) before validating those stats, considering current statsstore flushes those StatGenerator stats per 10s. I'm inclined not to add sleep(10s) there.

Why do we care about flushing? Can't you just read the current state from the store?

freedomljc:
> Why do we care about flushing? Can't you just read the current state from the store?

The stats from [StatGenerator](https://github.com/lyft/gostats/blob/118d7f89597e39f7e3c9e2b7539bb86960f6e19d/stats.go#L174)(including new-added [local cache stats](https://github.com/lyft/ratelimit/blob/09d62da657f1ccb98125737adf6c0d1df1b2d099/src/redis/local_cache_stats.go)) are different, and those stats are saved to the store while flushing.
For our case, the new-added local cache stats are saved to the store in the [GenerateStats](https://github.com/lyft/ratelimit/blob/09d62da657f1ccb98125737adf6c0d1df1b2d099/src/redis/local_cache_stats.go#L34) method, and the method `GenerateStats` gets invoked in the [Flush](https://github.com/lyft/gostats/blob/118d7f89597e39f7e3c9e2b7539bb86960f6e19d/stats.go#L309) method. The current statsstore invoked the Flush method every [10s](https://github.com/lyft/gostats/blob/118d7f89597e39f7e3c9e2b7539bb86960f6e19d/stats.go#L192), that's why we'll have to wait for 10s before validating the new-added [local cache stats](https://github.com/lyft/ratelimit/blob/09d62da657f1ccb98125737adf6c0d1df1b2d099/src/redis/local_cache_stats.go).

mattklein123:
OK that makes sense, but why can't you just manually flush the store before checking the stats?

freedomljc:
> OK that makes sense, but why can't you just manually flush the store before checking the stats?

Sorry, didn't realize that before even though i did the trick in the [unit test](https://github.com/lyft/ratelimit/blob/893d04c748a38eca424944fa0f25da1e14b141d0/test/redis/cache_impl_test.go#L160). Just added it.

test/integration/runtime/current/ratelimit/config/another_legacy.yaml
```diff
@@ -6,16 +6,6 @@ descriptors:
...
-    rate_limit:
-      unit: hour
-      requests_per_unit: 10
-
-  - key: key2_local
-    rate_limit:
-      unit: minute
-      requests_per_unit: 20
-
-  - key: key3_local
```
mattklein123: Sorry why are these removed here and below?

freedomljc: Because we just made the change to undo the local cache test in TestBasicConfigLegacy (https://github.com/lyft/ratelimit/pull/114#issuecomment-574914116), i removed the related data as well.

mattklein123: OK got it, thanks.

PR #115, opened 2020-01-06
Plugin statstore into runner

freedomljc:
Add statsstore as one parameter of runner, then be able to validate the internal stats inside the integration test.


freedomljc:
🔨

test/integration/integration_test.go
```diff
@@ -44,15 +46,15 @@ func newDescriptorStatusLegacy(
 func TestBasicConfig(t *testing.T) {
 	t.Run("WithoutPerSecondRedis", testBasicConfig("8083", "false", "0"))
 	t.Run("WithPerSecondRedis", testBasicConfig("8085", "true", "0"))
-	t.Run("WithoutPerSecondRedisWithLocalCache", testBasicConfig("8083", "false", "1000"))
-	t.Run("WithPerSecondRedisWithLocalCache", testBasicConfig("8085", "true", "1000"))
+	t.Run("WithoutPerSecondRedisWithLocalCache", testBasicConfig("18083", "false", "1000"))
```
freedomljc: Changing the port, to avoid the statstore gets reused among those tests. It is related to the go routine(https://github.com/lyft/ratelimit/blob/2c5f42aaaa74d5818829ff1473a915edda3e456c/test/integration/integration_test.go#L222-L224) won't exit after the method ends.

freedomljc:
@mattklein123 , can you take a look again?  Thanks.
And also found another separate issue https://github.com/lyft/ratelimit/issues/118.

test/integration/integration_test.go
mattklein123: OK. Can you add a TODO and comment about this? We should fix this so everything gets cleaned up after each test method but we can do that later.

test/integration/integration_test.go
```diff
@@ -93,9 +95,10 @@ func testBasicBaseConfig(grpcPort, perSecond string, local_cache_size string) fu
 
 		local_cache_size_val, _ := strconv.Atoi(local_cache_size)
 		enable_local_cache := local_cache_size_val > 0
+		store := stats.NewDefaultStore()
 
 		go func() {
-			runner.Run()
+			runner.Run(store)
```
mattklein123: I think it would be a bit cleaner to do something like:
1) Allocate a runner object
2) Have a method to get the store
3) Then have the run method which is a method on the object and not static.

This would avoid special casing and allow the test code to grab the store.

freedomljc: thanks for the suggestion. Just changed it, it looks better.

test/integration/integration_test.go
freedomljc: just added TODO, and also created a github issue on it.

PR #116, opened 2020-01-07
fix: support auth without tls

tangxinfa:
Redis's auth is not related with SSL, a non-ssl connection can have auth too.

### DIFF
--- a/src/redis/driver_impl.go
+++ b/src/redis/driver_impl.go
@@ -2,6 +2,7 @@ package redis
 
 import (
 	"crypto/tls"
+	"net"
 
 	stats "github.com/lyft/gostats"
 	"github.com/lyft/ratelimit/src/assert"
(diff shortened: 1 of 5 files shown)


tangxinfa:
@junr03 @mattklein123

This pull request already passed the tests, please have a review, thanks:)

src/redis/driver_impl.go
```diff
@@ -103,6 +94,29 @@ func NewAuthTLSPoolImpl(scope stats.Scope, auth string, url string, poolSize int
 		stats: newPoolStats(scope)}
 }
 
+func NewAuthPoolImpl(scope stats.Scope, auth string, url string, poolSize int) Pool {
```
mattklein123: This is basically the same function as the one above. Can you just merge it all into a single creation function that decides whether to use TLS, auth, etc. independently?

tangxinfa: done

mattklein123:
Can you merge master and re-run the integration tests? I think the ports might need to be changed per https://github.com/lyft/ratelimit/pull/115. cc @freedomljc

tangxinfa:
Already rebased with the latest master branch, all commits merged. Please review again

@mattklein123

Issue #129, opened 2020-03-24
proposal: radix.v2 to radix.v3


caitong93:
**Description:**

As [radix.v2](https://github.com/mediocregopher/radix.v2) is deprecated for a time, upgrading to [radix.v3](https://github.com/mediocregopher/radix)  give benefits on performance and maintenance.

The biggest change will be [Pool](https://github.com/envoyproxy/ratelimit/blob/master/src/redis/driver.go#L11) interface . [Pool](https://github.com/mediocregopher/radix/blob/master/pool.go#L238)  in radix.v3 implements implicit pipelining,  no need to explicitly `Get` and `Put` connections from pool.

**Changes:**
- intrefaces: RateLimitCache, Pool ..
- server initialization
- unit and integration tests, mocks

I'm glad to write a POC pr if we think it's worth doing.

mattklein123:
Sure PRs appreciated. Thank you!

Issue #133, opened 2020-04-12
Using Redis in socket mode no longer works


moderation:
I had a demo of ratelimit running in a Kubernetes cluster at the end of 2019. For performance reasons I had ratelimit connect to a Redis instance over a socket shared on a ConfigMap volume. Having the ratelimit -> Redis connection not use a network at all is great for performance.

Turns out https://github.com/envoyproxy/ratelimit/commit/3ec0f5f3a660ab3ce1f6e16f3823e18dd1ceebf7#diff-ab160e3a4ff5cb5fd488f666c5266fdd disabled the default Unix socket usage by hardcoding the use of TCP.

It is not possible to launch ratelimit with `REDIS_SOCKET_TYPE=unix` even though it is the 'default' value. See an example below:

```
env USE_STATSD=false LOG_LEVEL=debug REDIS_URL=/tmp/redis.sock REDIS_SOCKET_TYPE=unix RUNTIME_ROOT=examples/ratelimit/config/ RUNTIME_SUBDIRECTORY=. bin/ratelimit
WARN[0000] statsd is not in use
DEBU[0000] runtime changed. loading new snapshot at examples/ratelimit/config
DEBU[0000] runtime: processing examples/ratelimit/config
DEBU[0000] runtime: processing examples/ratelimit/config/config.yaml
DEBU[0000] runtime: adding key=config.yaml value=---
domain: mongo_cps
descriptors:
  - key: database
    value: users
    rate_limit:
      unit: second
      requests_per_unit: 500

  - key: database
    value: default
    rate_limit:
      unit: second
      requests_per_unit: 500
 uint=false
WARN[0000] connecting to redis on /tmp/redis.sock with pool size 10
panic: dial tcp: address /tmp/redis.sock: missing port in address

goroutine 1 [running]:
github.com/envoyproxy/ratelimit/src/redis.checkError(...)
        /home/moderation/Library/envoyproxy/ratelimit/src/redis/driver_impl.go:44
github.com/envoyproxy/ratelimit/src/redis.NewPoolImpl(0xca07e0, 0xc00010eb40, 0x0, 0x0, 0x0, 0xc00003a2ea, 0xf, 0xa, 0xc00011e080, 0xc000134000)
        /home/moderation/Library/envoyproxy/ratelimit/src/redis/driver_impl.go:98 +0x344
github.com/envoyproxy/ratelimit/src/service_cmd/runner.(*Runner).Run(0xc000213f68)
[...]

PR #137, opened 2020-05-20
Upgrade radix

caitong93:
issue: https://github.com/envoyproxy/ratelimit/issues/129

Changes:
- Replace `Pool` with `Client`, as in radix `Client` is abstract interface for pool, sentinel and redis cluster.
- Refactor redis cache impl to use implicit pipelining to improve throughput.

cc @mattklein123


src/redis/cache_impl.go
```diff
@@ -18,16 +18,16 @@ import (
...
+	// If this client is nil, then the Cache will use the client for all
+	// limits regardless of unit. If this client is not nil, then it
 	// is used for limits that have a SECOND unit.
-	perSecondPool              Pool
+	perSecondClient            Client
 	timeSource                 TimeSource
 	jitterRand                 *rand.Rand
 	expirationJitterMaxSeconds int64
-	// bytes.Buffer pool used to efficiently generate cache keys.
+	// bytes.Buffer client used to efficiently generate cache keys.
```
mattklein123, line 30: I think this is an errant find/replace

src/redis/driver_impl.go
```diff
@@ -45,78 +49,53 @@ func checkError(err error) {
...
-		return client, nil
+
+		return radix.Dial(network, addr, dialOpts...)
 	}
-	pool, err := pool.NewCustom("tcp", url, poolSize, df)
+
+	stats := newPoolStats(scope)
+
+	// TODO: support sentinel and redis cluster
+	pool, err := radix.NewPool("tcp", url, poolSize, radix.PoolConnFunc(df),
```
mattklein123, line 79: cc @moderation I think there was some regression here where we no longer support unix domain sockets, is that right? If so can we fix that also either here or in a follow up?

src/redis/cache_impl.go
caitong93: fixed

caitong93:
> Do any documentation need to be updated about supported config variables on the README? Can you check?

description about new config variables have been updated in README

src/settings/settings.go
```diff
@@ -9,29 +11,33 @@ type Settings struct {
...
+	RuntimePath                  string        `envconfig:"RUNTIME_ROOT" default:"/srv/runtime_data/current"`
+	RuntimeSubdirectory          string        `envconfig:"RUNTIME_SUBDIRECTORY"`
+	RuntimeIgnoreDotFiles        bool          `envconfig:"RUNTIME_IGNOREDOTFILES" default:"false"`
+	LogLevel                     string        `envconfig:"LOG_LEVEL" default:"WARN"`
+	RedisSocketType              string        `envconfig:"REDIS_SOCKET_TYPE" default:"unix"`
+	RedisUrl                     string        `envconfig:"REDIS_URL" default:"/var/run/nutcracker/ratelimit.sock"`
+	RedisPoolSize                int           `envconfig:"REDIS_POOL_SIZE" default:"10"`
+	RedisAuth                    string        `envconfig:"REDIS_AUTH" default:""`
+	RedisTls                     bool          `envconfig:"REDIS_TLS" default:"false"`
+	RedisPipelineWindow          time.Duration `envconfig:"REDIS_PIPELINE_WINDOW" default:"75µs"`
```
caitong93, line 29: Defaults for pipeline window and limit are chosen according to my e2e benchmark results. Also have a look at [this issue](https://github.com/mediocregopher/radix/issues/147).

I set up a 8c8g envoy with 10 2c2g ratelimit servers, and got 60000 requests/second . `75us` window seems better than `150us`(radix default) in this load.

src/redis/cache_impl.go
```diff
@@ -18,12 +18,12 @@ import (
 )
 
 type rateLimitCacheImpl struct {
-	pool Pool
-	// Optional Pool for a dedicated cache of per second limits.
-	// If this pool is nil, then the Cache will use the pool for all
-	// limits regardless of unit. If this pool is not nil, then it
+	client Client
+	// Optional Client for a dedicated cache of per second limits.
+	// If this client is nil, then the Cache will use the client for all
```
freedomljc, line 23: Might be not related to your change. 
But the description seems not right regarding `if this client is nil`. I think we need to specify it as at least a pointer, then we can determine whether it is nil or not.

src/redis/cache_impl.go
```diff
@@ -154,6 +141,8 @@ func (this *rateLimitCacheImpl) DoLimit(
 	}
 
 	isOverLimitWithLocalCache := make([]bool, len(request.Descriptors))
+	results := make([]uint32, len(request.Descriptors), len(request.Descriptors))
```
freedomljc, line 144: Is it intended? can you let us know why we need to specify `make([]uint32, len(request.Descriptors), len(request.Descriptors))` instead of `make([]uint32, len(request.Descriptors))`

src/redis/cache_impl.go
```diff
@@ -178,13 +167,20 @@ func (this *rateLimitCacheImpl) DoLimit(
 			expirationSeconds += this.jitterRand.Int63n(this.expirationJitterMaxSeconds)
 		}
 
+		result := &results[i]
+		key := cacheKey.key
```
freedomljc, line 171: I'd suggest not to add new variables `result` and `key` here, as the existing variables `results[i]` and `cacheKey.key` are already obvious to understand.

src/redis/cache_impl.go
```diff
@@ -178,13 +167,20 @@ func (this *rateLimitCacheImpl) DoLimit(
...
+				lastErr = err
+			}
 		} else {
-			pipelineAppend(conn, cacheKey.key, hitsAddend, expirationSeconds)
+			if err := pipelineAppend(this.client, key, hitsAddend, result, expirationSeconds); err != nil {
+				lastErr = err
+			}
 		}
 	}
+	checkError(lastErr)
```
freedomljc, line 183: can you let us know why only to panic in last error, instead of panic in the first error?

src/redis/driver_impl.go
```diff
@@ -45,78 +49,53 @@ func checkError(err error) {
...
-func NewPoolImpl(scope stats.Scope, useTls bool, auth string, url string, poolSize int) Pool {
-	logger.Warnf("connecting to redis on %s with pool size %d", url, poolSize)
-	df := func(network, addr string) (*redis.Client, error) {
-		var conn net.Conn
 		var err error
 		if useTls {
-			conn, err = tls.Dial("tcp", addr, &tls.Config{})
-		} else {
-			conn, err = net.Dial("tcp", addr)
+			dialOpts = append(dialOpts, radix.DialUseTLS(&tls.Config{}))
```
freedomljc, line 61: might be a dumb question. For non-tls, non-auth case, the dialOpts would be empty, is it still able to connect to redis?

src/redis/driver_impl.go
```diff
@@ -45,78 +49,53 @@ func checkError(err error) {
...
-	pool, err := pool.NewCustom("tcp", url, poolSize, df)
+
+	stats := newPoolStats(scope)
+
+	// TODO: support sentinel and redis cluster
+	pool, err := radix.NewPool("tcp", url, poolSize, radix.PoolConnFunc(df),
+		radix.PoolPipelineWindow(pipelineWindow, pipelineLimit),
+		radix.PoolWithTrace(poolTrace(&stats)),
+	)
 	checkError(err)
```
freedomljc, line 83: Can you add some test cases for the constructor `NewClientImpl `? especially when there exist some errors.

src/redis/driver_impl.go
```diff
@@ -45,78 +49,53 @@ func checkError(err error) {
...
+	return &clientImpl{
+		client: pool,
+		stats:  stats,
+	}
 }
 
-func (this *connectionImpl) PipeResponse() Response {
-	assert.Assert(this.pending > 0)
-	this.pending--
+func (c *clientImpl) DoCmd(rcv interface{}, cmd, key string, args ...interface{}) error {
```
freedomljc, line 91: Same here, it would be great if we have some tests to cover the error cases.

src/redis/driver_impl.go
moderation: Here is the issue for the socket regression where the functionality was completed removed in a prior commit - https://github.com/envoyproxy/ratelimit/issues/133. I unblocked myself but I think reintroduction will require a bit more work. Tests lacking if it was possible to delete the default documented working mode

src/redis/cache_impl.go
caitong93: It is an interface, should be OK for nil testing.

src/redis/driver_impl.go
caitong93: Yes.

src/redis/cache_impl.go
caitong93: When wrote this code, I was thinking to execute all commands  and recording all errors. For simplicity, only lastErr was checked. But now thinking again, it seems check first error is enough. I will update this latter.

src/redis/cache_impl.go
caitong93: will do

src/redis/cache_impl.go
caitong93: No, it's not intended. Will remove the redundant.

caitong93:
@freedomljc  have fixed comments and added some unit tests(just cover some basic cases).
Can you take a look again? Thanks :)

test/redis/cache_impl_test.go
```diff
@@ -529,3 +420,107 @@ func TestRedisWithJitter(t *testing.T) {
...
+
+	mkRedisClient := func(auth, addr string) redis.Client {
+		return redis.NewClientImpl(statsStore, false, auth, addr, 1, 1*time.Millisecond, 1)
+	}
+
+	t.Run("no connection", func(t *testing.T) {
+		assert.Panics(t, func() {
+			// It's possible there is a redis server listening on 6379 in ci environment, so
+			// use a random port.
+			mkRedisClient("", "localhost:12345")
```
freedomljc, line 445: Is it to test the [ping/pong connection](https://github.com/envoyproxy/ratelimit/blob/d08853c479faffb87b799e28baf4580e9159eb48/src/redis/driver_impl.go#L87-L90) ? can we also check the error?

test/redis/cache_impl_test.go
```diff
@@ -529,3 +420,107 @@ func TestRedisWithJitter(t *testing.T) {
...
+	t.Run("auth pass", func(t *testing.T) {
+		redisSrv := mustNewRedisServer()
+		defer redisSrv.Close()
+
+		redisSrv.RequireAuth(redisAuth)
+
+		assert.NotPanics(t, func() {
+			mkRedisClient(redisAuth, redisSrv.Addr())
+		})
+	})
```
freedomljc, line 480: is it possible to test tls case as well?

caitong93: tls support is not available for now. Maybe we can do it in another pr when 
https://github.com/alicebob/miniredis/issues/155 solved ?

test/redis/cache_impl_test.go
caitong93: It is to test connection refused case. `ping/pong` is used to check auth.(we don't know if auth is ok util receive first response from redis).
Sure we can check the error. It require update testify package.

test/redis/bench_test.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+	}
+
+	b.Run("no pipeline", mkDoLimitBench(0, 0))
+	b.Run("pipeline 50us 4", mkDoLimitBench(50*time.Microsecond, 4))
+	b.Run("pipeline 75us  4", mkDoLimitBench(75*time.Microsecond, 4))
+	b.Run("pipeline 150us 4", mkDoLimitBench(150*time.Microsecond, 4))
+	b.Run("pipeline 50us 8", mkDoLimitBench(50*time.Microsecond, 8))
+	b.Run("pipeline 75us 8", mkDoLimitBench(75*time.Microsecond, 8))
+	b.Run("pipeline 150us 8", mkDoLimitBench(150*time.Microsecond, 8))
```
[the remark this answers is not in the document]
caitong93: I compared `4` , `8` and `16` in e2e benchmark, using tcpdump to observe actual pipeline size. Found many requests to redis have a pipeline size greater than `4`, nearly `8`. This analysis is very rough, but `8` should be ok to use in most situations.

caitong93:
@freedomljc Updated

test/redis/cache_impl_test.go
freedomljc: okay, sound fair.

caitong93:
> One more suggestion: pls do not force-push the pr, as it loses track of the context among commits/comments.

Thanks the suggestion ! ( always forgot DCO and use force-push to fix )

Conflicts resolved, PTAL  @freedomljc @mattklein123

PR #141, opened 2020-05-26
Upgrade gostats dependency from 0.2.6 to 0.4.0

dweitzman:
My interest is the UDP protocol support which appeared in gotstats 0.3.10

There's a breaking change as of https://github.com/lyft/gostats/releases/tag/v0.3.0
which is that gostats no longer publishes stats as expvars.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -18,8 +18,6 @@ github.com/gogo/protobuf v1.3.1/go.mod h1:SlYgWuQ5SjCEi6WLHjHCa1yvBfUnHcTbrrZtXP
 github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b h1:VKtxabqXZkF25pY9ekfRL6a582T4P37/31XEstQ5p58=
 github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b/go.mod h1:SBH7ygxi8pfUlaOkMMuAQtPIUF8ecWP5IEl/CR7VP2Q=
 github.com/golang/mock v1.1.1/go.mod h1:oTYuIxOrZwtPieC+H1uAHpcLFnEyAGVDL/k47Jfbm0A=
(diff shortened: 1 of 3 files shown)


mattklein123:
Looks like CI is broken? Can you check?

dweitzman:
Ah, I see what happened. https://github.com/lyft/gostats/pull/89 flipped the default for USE_STATSD from false to true, so the integration tests were timing out trying to contact an external statsd service which was not actually running

Updated the test, which will hopefully pass now...

mattklein123:
@dweitzman please don't force push as it makes reviews much more difficult. Just add commits and merge master. Thank you!

go.sum
```diff
@@ -39,6 +39,8 @@ github.com/lyft/goruntime v0.2.1 h1:7DebA8oMVuoQ5TQ0j1xR/X2xRagbGrm0e2SoMdt5tRs=
 github.com/lyft/goruntime v0.2.1/go.mod h1:8rUh5gwIPQtyIkIXHbLN1j45HOb8cMgDhrw5GA7DF4g=
 github.com/lyft/gostats v0.2.6 h1:m4XmqpBamBXaFjp76h2Ao4TrNpsIVODNClDrH0YTbjM=
 github.com/lyft/gostats v0.2.6/go.mod h1:Tpx2xRzz4t+T2Tx0xdVgIoBdR2UMVz+dKnE3X01XSd8=
+github.com/lyft/gostats v0.4.0 h1:PbRWmwidTPk6Y80S6itBWDa+XVt1hGvqFM88TBJYdOo=
```
mattklein123, line 42: Shouldn't the lines above be removed?

dweitzman: Updated by running `go mod tidy` which removed the unused entries from go.sum

mattklein123:
@dweitzman please never force push. Thank you!

Issue #140, opened 2020-05-26
Proposal to contribute memcache support in addition to redis


dweitzman:
We're interested in contributing memcache support as an optional alternative to redis for the ratelimit k/v store. Memcache is what we normally use for caching, and it's what our storage & caching team is comfortable with and provides as a service to us.
Does this sound like a totally crazy and useless contribution that would be rejected if we mailed a clean and clear PR?

mattklein123:
> Does this sound like a totally crazy and useless contribution that would be rejected if we mailed a clean and clear PR?

Sure sounds fine to me. IIRC the cache abstraction should make it relatively easy to do this.

PR #142, opened 2020-05-27
Split redis-specific logic from generic key-value store logic

dweitzman:
This is a pure refactoring with no behavior changes. It's a step toward being able
to add memcache as a backend (see #140).

This PR moves RateLimitCache from the redis package to a new "limiter" package, along with
code for time/jitter and for constructing cache keys. All that can be reused with memcache.

After this PR, the redis package is imported in exactly two places:
- in service_cmd/runner/runner.go to call redis.NewRateLimiterCacheImplFromSettings()
- in service/ratelimit.go in ShouldRateLimit to identify if a recovered panic is a redis.RedisError. If so, a stat is incremented and the panic() propagation is ended and in favor of returning the error as a the function result.

The PR also includes changes by goimports to test/service/ratelimit_test.go
so that the difference between package name vs file path name is explicit
instead of implicit.


test/mocks/redis/redis.go
```diff
@@ -1,227 +1,151 @@
-// Automatically generated by MockGen. DO NOT EDIT!
-// Source: github.com/envoyproxy/ratelimit/src/redis (interfaces: RateLimitCache,Pool,Connection,Response,TimeSource,JitterRandSource)
+// Code generated by MockGen. DO NOT EDIT.
```
dweitzman, line 1: Looks like this text changed in mockgen 3 years ago (https://github.com/golang/mock/commit/c9cc560280e96593c64087bea69b082fd7460156) prior to the v1.0.0 mockgen release. I may send out a follow-up to try to make sure everyone is using the same mockgen version even if they're not using `make bootstrap`.

mattklein123:
Please merge master.

dweitzman:
> Please merge master.

Done

mattklein123:
Apologies but this is going to need a big merge with master. The other PR got started first so I merged it first.

dweitzman:
Rebased

mattklein123:
@dweitzman please *never* force push as it breaks reviews. Just add commits and merge master. Thanks!

dweitzman:
> @dweitzman please _never_ force push as it breaks reviews. Just add commits and merge master. Thanks!

Noted and apologies. For this PR Would it be helpful at this point to change it back to a merge from the previous commit instead of a rebase?

mattklein123:
> Noted and apologies. For this PR Would it be helpful at this point to change it back to a merge from the previous commit instead of a rebase?

It's fine as I haven't reviewed yet.

PR #146, opened 2020-06-08
WIP: add ip whitelist and forceflag

nicholasxuu:
- Add IP whitelist setting
- add forceFlag setting to force turn off rate limit


mattklein123:
Please open an issue describing what feature you are trying to add before opening the feature PR. Thank you!

PR #148, opened 2020-06-10
json handler: return full ratelimit service response as json

dweitzman:
Previously an HTTP POST to /json would only return an HTTP status code,
not all the other details supported by grpc ratelimit responses.

With this change an HTTP POST to /json receives the full proto3 response
encoded as json by jsonpb.

It seems unlikely that anyone would be parsing the text "over limit" from
the HTTP body instead of just reading the 429 response code, but for anyone
doing that this would be a breaking change.


src/server/server_impl.go
```diff
@@ -73,11 +76,20 @@ func (server *server) AddJsonHandler(svc pb.RateLimitServiceServer) {
...
 		if resp.OverallCode == pb.RateLimitResponse_OVER_LIMIT {
-			http.Error(writer, "over limit", http.StatusTooManyRequests)
+			writer.WriteHeader(http.StatusTooManyRequests)
 		} else if resp.OverallCode == pb.RateLimitResponse_UNKNOWN {
-			http.Error(writer, "unknown", http.StatusInternalServerError)
+			writer.WriteHeader(http.StatusInternalServerError)
+		}
+
+		err = m.Marshal(writer, resp)
+		if err != nil {
```
mattklein123, line 90: Can this actually happen? If not can you assert it instead?

dweitzman: My _guess_ is that there probably couldn't be a serialization error but probably could be an I/O error if the remote peer closes the TCP connection, in which case the fmt.Fprintf() would fail anyway.

I could split the serialization from the writing to the socket and do a assert/panic and recover dance to print the stack trace if a serialization error happens. If that's the route, I'd kind of rather just do a normal log.Error() instead of an assert because I don't think the extra code to deal with panic()s to get a stack trace will provide much value.

Up to you.

mattklein123: I'm fine either way, I mostly just want to avoid error handling and just crash if there are things that should never happen.

dweitzman:
> Given this was just added, I think we can probably change this, and agreed it makes sense. Though I do wonder if we should still keep the HTTP response code as a 429 in the over limit case? I don't feel strongly about it.

This PR leaves the status codes the same. The integration tests verify that.

> 
> Also, can you update [/README.md@`master`#json-endpoint](https://github.com/envoyproxy/ratelimit/blob/master/README.md#json-endpoint)?
> 

Done

> Thank you!

src/server/server_impl.go
dweitzman: My sense is:

- handling the error differently than how it's handled here is probably more code
- I'd be comfortable leaving the code here as-is and waiting until someone has evidence that a json serialization error can actually happen before writing additional logic to handle it in a more nuanced way

More generally on the topic of when to crash the process, it seems like crashing a process makes sense when:
- The process is permanently severely degraded or broken
- Restarting has a decent chance of un-wedging it

My sense is that a single unknown json serialization error in isolation probably doesn't justify a crash, vs something like a resource allocation issue potentially could justify a crash.

dweitzman:
Had to force push due to missing DCO in the new commit

src/server/server_impl.go
mattklein123: The guiding principle is whether you can test this or not. I don't think you can test this branch, as such I would use https://github.com/envoyproxy/ratelimit/blob/master/src/assert/assert.go and not handle the error.

dweitzman: Added unit tests for all the cases. For the serialization failure I used a 'nil' return value from ShouldRateLimit(). A non-nil serialization problem would theoretically be possible if the ShouldRateLimit() API ever evolves to include `Any` types, but I don't think there are `Any` types in the spec today

mattklein123: Sounds good, thanks. Will take a look.

test/server/server_impl_test.go
```diff
@@ -0,0 +1,86 @@
...
+	rls.EXPECT().ShouldRateLimit(nil, &pb.RateLimitRequest{
+		Domain: "foo",
+	}).Return(&pb.RateLimitResponse{}, nil)
+	assertHttpResponse(t, handler, `{"domain": "foo"}`, 500, "application/json", "{}")
+
+	// ratelimit service error
+	rls.EXPECT().ShouldRateLimit(nil, &pb.RateLimitRequest{
+		Domain: "foo",
+	}).Return(nil, fmt.Errorf("some error"))
+	assertHttpResponse(t, handler, `{"domain": "foo"}`, 400, "text/plain; charset=utf-8", "some error\n")
```
dweitzman, line 63: Note: I kinda think this sort of error from ShouldRateLimit should result in a 500 rather than a 400, but changing that is not in scope for this PR.

mattklein123:
Please fix DCO.

/wait

dweitzman:
> Please fix DCO.
> 
> /wait

Fixed

PR #155, opened 2020-07-02
Followups to v3 upgrade

Pchelolo:
- Regenerate mocks based on new default protocol
- Manually transform v2 messages to v3 messages - some of the fields
were renamed thus json Marshal/Unmarshal does not work anymore
- Added tests that verify conversion v2<->v3 works for headers fields
- Update tests to use proto.Equal - simple assert.Equals might not
work correctly for protobuf messages stretchr/testify#758

Signed-off-by: Petr Pchelko ppchelko@wikimedia.org


Pchelolo:
ping for review

PR #156, opened 2020-07-08
Introduce a Dockerfile for running integration tests

dweitzman:
This diff creates Dockerfile.integration for running integration tests with clearly-defined dependencies. Previously the dependencies of the integration tests were defined within the github actions config.

The new "make docker_tests" target should work for any developer with Docker installed.

Previously there was no single command that would run integration tests across platforms, which makes development and onboarding harder. Even copying the command from github actions wouldn't have worked before, since that command quietly assumed that redis was already running on port 6379.

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -17,18 +17,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: deps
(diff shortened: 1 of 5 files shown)


ysawa0:
Works on my fork and locally. I couldn't figure out how to run integration tests locally before so this is great.

PR #157, opened 2020-07-09
Allow statsd protocol to be configurable

ianbishop:
This capability is already available via [lyft/gostats](https://github.com/lyft/gostats/blob/cf86465edce66ad764b7412ea3dda2800c0bccdc/settings.go#L32), this just exposes it to be configurable in ratelimit. We require this as we only support UDP statsd for performance reasons.

Signed-off-by: Ian Bishop <ibishop@flipboard.com>

### DIFF
--- a/src/settings/settings.go
+++ b/src/settings/settings.go
@@ -17,6 +17,7 @@ type Settings struct {
 	UseStatsd                    bool          `envconfig:"USE_STATSD" default:"true"`
 	StatsdHost                   string        `envconfig:"STATSD_HOST" default:"localhost"`
 	StatsdPort                   int           `envconfig:"STATSD_PORT" default:"8125"`
+	StatsdProtocol               string        `envconfig:"STATSD_PROTOCOL" default:"tcp"`
 	RuntimePath                  string        `envconfig:"RUNTIME_ROOT" default:"/srv/runtime_data/current"`
 	RuntimeSubdirectory          string        `envconfig:"RUNTIME_SUBDIRECTORY"`
 	RuntimeIgnoreDotFiles        bool          `envconfig:"RUNTIME_IGNOREDOTFILES" default:"false"`


src/settings/settings.go
```diff
@@ -17,6 +17,7 @@ type Settings struct {
 	UseStatsd                    bool          `envconfig:"USE_STATSD" default:"true"`
 	StatsdHost                   string        `envconfig:"STATSD_HOST" default:"localhost"`
 	StatsdPort                   int           `envconfig:"STATSD_PORT" default:"8125"`
+	StatsdProtocol               string        `envconfig:"STATSD_PROTOCOL" default:"tcp"`
```
dweitzman, line 20: Does setting STATSD_PROTOCOL without making this change not work for you?

(For me it seems to work without this change, just with the declaration in https://github.com/lyft/gostats/blob/f9add79ec5fce9579fb2e52b6b419da2ee965b37/settings.go#L26)

dweitzman: Although even if it works without this, declaring it here might be useful from a documentation / discoverability perspective

ianbishop: I did try setting `STATSD_PROTOCOL` and thought that it did not work which led to this PR. I will try it again, I probably made an error.

ianbishop: Sorry, the issue seems to be that `STATS_PROTOCOL` does not work with the `envoyproxy/ratelimit:v1.4.0` container. Building a new container from main works without this change.

Would you like me to close it or leave it for documentation purposes?

dweitzman: I don't have strong options either way, but also I'm not a maintainer so I wouldn't be able to merge regardless

Seems like removing the statsd stuff (host, port, etc.) from settings.go and just linking from the README to whatever documentation exists in the gostats library would require the least amount of ongoing maintenance

mattklein123: +1 can we just move this to the README?

PR #158, opened 2020-07-14
Add support for rate limit overrides.

### LINKED ISSUE #154: Support rate limit overrides
A new feature was added in https://github.com/envoyproxy/envoy/issues/11595. In a nutshell it allows providing an override limit in the request. This is useful if the source of truth for limits is in some authentication service, so the limits can be encoded and supplied to the proxy via an authentication service. Support for that feature needs to be added to rate limiter service.

The semantics would be the following:
- If there's no limit override provided in the descriptor, everything behaves as is used to.
- If there's a limit override provided in the descriptor, we treat it as if there was a match for this descriptor in the static config and the configured limit is what's in the override
- If we had previously received a request with a limit override (there's a RateLimit struct exist for the descriptor key), and now the override value for the descriptor has changed, we update the limit override value but keep the stats. This will allow for eventual consistency of the rates. The values for the overrides are not expected to change on every request, so eventual consistency should be good enough here.

Supporting this requires having a new release of the go-control-plane, so I will submit a WIP PR, but will have to wait for a new version of the control plane to be cut.

Pchelolo:
Fixes #154

Signed-off-by: Petr Pchelko <ppchelko@wikimedia.org>


src/config/config_impl.go
```diff
@@ -252,6 +267,28 @@ func (this *rateLimitConfigImpl) GetLimit(
 		return rateLimit
 	}
 
+	if descriptor.GetLimit() != nil {
+		rateLimitKey := domain + "." + this.descriptorToKey(descriptor)
+		rateLimit := value.overrides[rateLimitKey]
```
mattklein123, line 272: Is this caching for perf? If so can you add more comments?

src/config/config_impl.go
```diff
@@ -252,6 +267,28 @@ func (this *rateLimitConfigImpl) GetLimit(
...
+		rateLimitOverrideUnit := pb.RateLimitResponse_RateLimit_Unit(descriptor.GetLimit().GetUnit())
+		if rateLimit != nil {
+			if rateLimit.Limit.RequestsPerUnit != descriptor.GetLimit().GetRequestsPerUnit() ||
+				rateLimit.Limit.Unit != rateLimitOverrideUnit {
+
+				rateLimit.Limit.RequestsPerUnit = descriptor.GetLimit().GetRequestsPerUnit()
+				rateLimit.Limit.Unit = rateLimitOverrideUnit
```
mattklein123, line 279: Aren't there potential concurrency issues here and below? I'm wondering if this is premature optimization and we should just allocate a new limit and return it to avoid potential issues?

Pchelolo: Ok, makes sense, thank you for the review. Removed the caching for now, in case we discover perf issues with allocating new structs every time I'll come back with a followup.

src/config/config_impl.go
Pchelolo: Dropped it

PR #160, opened 2020-07-24
Add stats gauge for currently configured limit

ikonst:
Adding a stats gauge for the currently configured limit.

Since the counters aggregate per minute, the gauge's key includes the unit (`limit_per_<unit>`) to ensure correct unit is being used when querying and alarming on capacity; that is, rather than `ts(.total_hits) / ts(.limit) > 0.8`, one should do `msum(1h, ts(.total_hits)) / ts(.limit_per_hour) > 0.8`.


ikonst:
On second thought, this doesn't work well. Ideally we either need a gauge for the current rate limit value and a gauge for the configured limit, or a single percentage gauge. We shouldn't rely on TS formulas to approximate the rate limiter.

mattklein123:
@ikonst can you more fully describe what you are trying to achieve? It's not completely clear to me.

ikonst:
The rationale is to be able to create dashboards and alarms around "capacity", something like what [DynamoDB](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/metrics-dimensions.html) provides in its `ConsumedReadCapacityUnits` and `ProvisionedReadCapacityUnits` metrics.

What I'm thinking now is adding `limit_utilized` (inverse of "limit remaining") and `limit` gauges. WDYT?

mattklein123:
> What I'm thinking now is adding limit_utilized (inverse of "limit remaining") and limit gauges. WDYT? (A different approach would have us add a single gauge for a percentage, but I think this is more flexible.)

This is fine with me, with the caveat that the gauge will be noisy and eventually consistent. If that is OK for you I think this makes sense.

mattklein123:
Looks like it can't be reopened. Feel free to open a new PR.

Issue #152, opened 2020-06-30
Switch default protocol to v3


Pchelolo:
**Rational**
In order to support newer features of the ratelimit protocol (like [rate limit overrides]( https://github.com/envoyproxy/envoy/issues/11595)), we’d need the service to use v3 protocol instead of v2 protocol. Thus, we propose to upgrade the ratelimiter core to v3 protocol. Since there doesn’t seem to be any real changes between v2 and v3, this seems to be as easy as replacing the imports and referencing newer versions in the data plane.

However, to support complete compatibility with v2, we’d need to have a legacy layer, just like it’s done for v1 now. Supporting 3 versions of the protocol with 2 legacies seems like an overkill, so we propose to simultaneously drop support for v1 ratelimit.proto. Seems like it should be ok given that the release schedule proposed doing so in [Q32018](https://github.com/envoyproxy/ratelimit#deprecation-schedule) [2]

**Proposal**
Thus, the concrete proposal:
Protocol v1 support is dropped
Protocol v2 becomes legacy and converted to v3. Given that the protocols are only different in message names, this should have almost no performance penalty
The service core is updated to use protocol v3

We intend to contribute all the code. We are also planning to use v3 protocol in production in our own fork in the beginning, thus we can provide some battle testing. What do you think?

mattklein123:
+1, awesome, thanks.

Pchelolo:
Resolved with #11595

Pchelolo:
I have realized a few followups are needed: #155

ysawa0:
@Pchelolo after testing the latest master with these changes I noticed a big performance degradation with V2 API. Under light load the p999 latency went from 15ms => 22ms which is not ideal as we try to keep p999 under 20ms to avoid as many fail-open rate limit requests.

I did a CPU profiling with pprof and saw there's significant time being spent in the converting of v2 request to v3:
https://github.com/envoyproxy/ratelimit/blob/master/src/service/ratelimit_legacy.go#L67-L93

Is there a better way we can convert the request? As v2 is to be officially deprecated at end of this year, moving to v3 is good (and we plan to move our Envoy configs to v3 as well) but I feel there are many orgs still on v2.

Pchelolo:
[...]

PR #163, opened 2020-08-07
redis client: default to use explicit pipelining

caitong93:
Fix https://github.com/envoyproxy/ratelimit/issues/164

See conversations in https://github.com/envoyproxy/ratelimit/issues/152#issuecomment-668212174

Before 137 , explicit pipelining was used. After 137, we change to implicit pipelining to improve throughput (sacrifice some latency). This can cause problems for users who are concerned about latency. This pr make an optimization for this. If user turn off implicit pipelining by configuring pipeline_window=0, pipeline_limit=0, the redis client will use explicit pipelining.

Here is benchmark stats for `DoLimit` , for pipeline_window=0, pipeline_limit=0 case

Comparing to 1657473ad4a344de5428ec98b494d5112ef242ff
```
benchstat /tmp/ratelimit-before-1657473a.stat /tmp/ratelimit-opt.stat
name                           old time/op    new time/op    delta
ParallelDoLimit/no_pipeline-4    12.9µs ± 0%     9.1µs ± 0%   ~     (p=1.000 n=1+1)

name                           old alloc/op   new alloc/op   delta
ParallelDoLimit/no_pipeline-4      624B ± 0%      408B ± 0%   ~     (p=1.000 n=1+1)

name                           old allocs/op  new allocs/op  delta
ParallelDoLimit/no_pipeline-4      29.0 ± 0%      20.0 ± 0%   ~     (p=1.000 n=1+1)
```

Comparing to master

```
benchstat /tmp/ratelimit-master.stat /tmp/ratelimit-opt.stat 
name                           old time/op    new time/op    delta
ParallelDoLimit/no_pipeline-4    16.1µs ± 0%     9.1µs ± 0%   ~     (p=1.000 n=1+1)

name                           old alloc/op   new alloc/op   delta
ParallelDoLimit/no_pipeline-4      302B ± 0%      408B ± 0%   ~     (p=1.000 n=1+1)

name                           old allocs/op  new allocs/op  delta
ParallelDoLimit/no_pipeline-4      16.0 ± 0%      20.0 ± 0%   ~     (p=1.000 n=1+1)
```

### DIFF
--- a/src/redis/driver.go
+++ b/src/redis/driver.go
@@ -19,7 +19,7 @@ type Client interface {
 	// @param args supplies the additional arguments.
 	DoCmd(rcv interface{}, cmd, key string, args ...interface{}) error
 
-	// Pipe append a command onto the pipeline queue.
+	// PipeAppend append a command onto the pipeline queue.
 	//
 	// @param pipeline supplies the queue for pending commands.
 	// @param rcv supplies receiver for the result.
@@ -41,6 +41,9 @@ type Client interface {
 
 	// NumActiveConns return number of active connections, used in testing.
 	NumActiveConns() int
+
+	// ImplicitPipeliningEnabled return true if implicit pipelining is enabled.
+	ImplicitPipeliningEnabled() bool
 }
 
 type Pipeline []radix.CmdAction
--- a/src/redis/driver.go
+++ b/src/redis/driver.go
@@ -1,5 +1,7 @@
 package redis
 
+import "github.com/mediocregopher/radix/v3"
+
 // Errors that may be raised during config parsing.
 type RedisError string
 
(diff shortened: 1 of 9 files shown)


mattklein123:
@caitong93 is this still WIP? If not can you change the title and fix DCO and then we can review?

src/redis/driver.go
```diff
@@ -17,10 +19,31 @@ type Client interface {
...
 	DoCmd(rcv interface{}, cmd, key string, args ...interface{}) error
 
+	// PipeAppend append a command onto the pipeline queue.
+	//
+	// @param pipeline supplies the queue for pending commands.
+	// @param rcv supplies receiver for the result.
+	// @param cmd supplies the command to append.
+	// @param key supplies the key to append.
+	// @param args supplies the additional arguments.
+	PipeAppend(pipeline Pipeline, rcv interface{}, cmd, key string, args ...interface{}) Pipeline
```
caitong93, line 29: `PipeAppend ` seems no need to be method of `Client` interface. Add it just for unit testing. Need more thinking.

caitong93:
@mattklein123 done, PTAL

Issue #161, opened 2020-07-27
docker-compose up failing: compatible with Docker Desktop for Mac?


leecalcote:
**System**
- Runtime: Docker Desktop for Mac v2.3.0.3 (stable channel)
- OS: macOS Catalina 10.15.6

I see the following failure when executing `docker-compose up` from the root of this repo:

```
➜  ratelimit git:(master) docker-compose up
Creating network "ratelimit_ratelimit-network" with the default driver
Creating network "ratelimit_default" with the default driver
Creating volume "ratelimit_binary" with default driver
Pulling redis (redis:alpine)...
alpine: Pulling from library/redis
df20fa9351a1: Already exists
9b8c029ceab5: Pull complete
e983a1eb737a: Pull complete
660ad543c5fc: Pull complete
823cbe4f5025: Pull complete
e3dd0c30e1c8: Pull complete
Digest: sha256:6972ee00fd35854dd2925904e23cb047faa004df27c62cba842248c7db6bd99d
Status: Downloaded newer image for redis:alpine
Pulling ratelimit-build (golang:1.14-alpine)...
1.14-alpine: Pulling from library/golang
df20fa9351a1: Already exists
ed8968b2872e: Pull complete
a92cc7c5fd73: Pull complete
00abe362642d: Pull complete
e0491c15e88f: Pull complete
Digest: sha256:d6deb50437547fd7226890aa19b075fbf7f5063e3c0d7a2eaf7fb41d11069013
Status: Downloaded newer image for golang:1.14-alpine
Pulling ratelimit (alpine:3.6)...
3.6: Pulling from library/alpine
5a3ea8efae5d: Pull complete
Digest: sha256:66790a2b79e1ea3e1dabac43990c54aca5d1ddf268d9a5a0285e4167c8b24475
Status: Downloaded newer image for alpine:3.6
Creating ratelimit_redis_1                  ... done
Creating ratelimit_ratelimit-build_1        ... done
Creating ratelimit_ratelimit-client-build_1 ... done
Creating ratelimit_ratelimit_1              ... error

ERROR: for ratelimit_ratelimit_1  Cannot start service ratelimit: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"/usr/local/bin/ratelimit\": stat /usr/local/bin/ratelimit: no such file or directory": unknown

[...]

PR #170, opened 2020-09-03
Add full test environment example. Fix bug in existing docker-compose.

ysawa0:
- Add docker-compose to set up a full test environment for demoing Envoy based rate limiting
- Fix bug in existing docker-compose as the ratelimit container would try running before the binary was built. #161

### DIFF
--- a/docker-compose-example.yml
+++ b/docker-compose-example.yml
@@ -1,0 +1,89 @@
+version: "3"
+services:
+  redis:
+    image: redis:alpine
+    expose:
+      - 6379
+    ports:
+      - 6379:6379
+    networks:
+      - ratelimit-network
+
+  statsd:
+    image: prom/statsd-exporter:v0.12.2
+    expose:
+      - 9125
+    ports:
+      - 9125:9125
+    networks:
+      - ratelimit-network
+
+  ratelimit:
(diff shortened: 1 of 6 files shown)


docker-compose-example.yml
```diff
@@ -0,0 +1,87 @@
...
+      - STATSD_PORT=9125
+      - LOG_LEVEL=debug
+      - REDIS_SOCKET_TYPE=tcp
+      - REDIS_URL=redis:6379
+      - RUNTIME_ROOT=/data
+      - RUNTIME_SUBDIRECTORY=ratelimit
+      - RUNTIME_WATCH_ROOT=false
+
+  envoy-proxy:
+    image: envoyproxy/envoy:v1.14.4
```
mattklein123, line 47: Thoughts on using the nightly Envoy image? It will make it more likely we keep things up to date? At the very least I would use v1.15.0? Same below.

mattklein123 (changes requested):
Thanks so much this is amazing!

docker-compose-example.yml
ysawa0: That makes sense. I will update it to envoy-dev:latest.

mattklein123 (approved):
Thanks!

PR #172, opened 2020-09-08
Implement BACKEND_TYPE=memcache as an alternative k/v store to redis

dweitzman:
MEMCACHE_HOST_PORT=host:port must be set with BACKEND_TYPE=memcache

To minimize roundtrips when getting multiple keys, the memcache implementation
does a GetMulti to fetch the existing rate limit usage and does increments
asynchronously in background goroutines, since the memcache API doesn't offer
multi-increment.

Resolves https://github.com/envoyproxy/ratelimit/issues/140


dweitzman:
I'm open to whatever sort of feedback or changes here. I wish there wasn't so much duplication of unit test cases between redis and memcache, though avoiding that cleanly seems tricky.

src/memcached/cache_impl.go
```diff
@@ -0,0 +1,300 @@
...
+	// Same as in limiter.RateLimitCache
+	DoLimit(
+		ctx context.Context,
+		request *pb.RateLimitRequest,
+		limits []*config.RateLimit) []*pb.RateLimitResponse_DescriptorStatus
+
+	// Waits for any lingering goroutines that are incrementing memcache values.
+	// This is used for unit tests, since the memcache increments happen
+	// asynchronously in the background.
+	Flush()
```
mattklein123, line 47: Would it be better to add the Flush() method to the main interface and just have it be a NOP for redis? I think that would be cleaner? Would this make it easier to share the tests between the two implementations?

mattklein123 (changes requested):
Thanks for working on this. I didn't review the memcached code in detail, but structurally this LGTM. I had one small comment about the interface. Also, can you update the docs/README to discuss this feature, and also take this out of draft when it's ready for final review?

nezdolik:
Hi @dweitzman @mattklein123 we at Spotify are also looking into memcache support for rate limiter service. Maybe we could sync effort for making this PR to be merged upstream?

dweitzman:
Hey @nezdolik. The reason I hadn't sent this out more officially was that I was thinking about whether there was a way to refactor to minimize duplication in the test logic across redis and memcache. It's not the end of the world to have some duplicated test logic, I suppose, so it might be better to just land it first and then try to refactor.

(Another thing I'd like to do at some point is have the test cases themselves start a redis/memcache instance rather than having a bunch of them pre-started on different ports)

We already using this memcache code in production for some services at Pinterest, incidentally.

@mattklein123 How do you feel about me mailing this is a non-draft as-is (after syncing and resolving merge conflicts) and then then trying in the future to see how practical it is to share test cases across the redis vs memcache tests?

mattklein123:
Sure I'm fine with that. @nezdolik can you help review? See also https://github.com/envoyproxy/ratelimit/pull/193 which may have implications on this that need to be sorted out.

dweitzman:
I updated the branch this PR was pointed at, but it seems like when a PR is "closed" a change to the branch doesn't seem to update it. Is it possible to re-open this PR somehow or should I create an entirely new one?

nezdolik:
Sure thing, I can help with review. Maybe @mattklein123 has permissions to reopen this PR?

mattklein123:
Done!

dweitzman:
I moved `CalculateReset` to the `utils` package so it could be shared across redis and memcache, and then to avoid a circular dependency I also moved `time.go` under `utils`.

dweitzman:
> Thanks for working on this. I didn't review the memcached code in detail, but structurally this LGTM. I had one small comment about the interface. Also, can you update the docs/README to discuss this feature, and also take this out of draft when it's ready for final review?

Updated readme

src/memcached/cache_impl.go
```diff
@@ -0,0 +1,294 @@
...
+
+	// Now fetch the pipeline.
+	responseDescriptorStatuses := make([]*pb.RateLimitResponse_DescriptorStatus,
+		len(request.Descriptors))
+
+	var memcacheValues map[string]*memcache.Item
+	var err error
+
+	if len(keysToGet) > 0 {
+		memcacheValues, err = this.client.GetMulti(keysToGet)
```
nezdolik, line 114: Should we document somewhere the limitation of max key length (250 bytes)? https://github.com/bradfitz/gomemcache/blob/a41fca850d0b6f392931a78cbae438803ea0b886/memcache/memcache.go#L454

src/memcached/cache_impl.go
```diff
@@ -0,0 +1,294 @@
...
+
+func (this *rateLimitMemcacheImpl) DoLimit(
+	ctx context.Context,
+	request *pb.RateLimitRequest,
+	limits []*config.RateLimit) []*pb.RateLimitResponse_DescriptorStatus {
+
+	logger.Debugf("starting cache lookup")
+
+	// request.HitsAddend could be 0 (default value) if not specified by the caller in the Ratelimit request.
+	hitsAddend := max(1, request.HitsAddend)
```
nezdolik, line 66: I see that bunch of code is duplicated across redis and memcache implementations particularly in this method. Would it be possible to factor that code our into some common method?

nezdolik (commented):
Will do one more pass tomorrow. Had some comments so far.

src/memcached/cache_impl.go
mattklein123: Note this comment is the same as this one: https://github.com/envoyproxy/ratelimit/pull/193#discussion_r532778253. It would be great to figure out some of this as I think the two PRs are going to run into very similar issues around duplication so I would take a look at my other comments in that one also.

nezdolik: ok, maybe we need a separate PR which pulls up duplicated code into single place

src/memcached/cache_impl.go
dweitzman: Added comments to cache_impl.go and README

src/memcached/cache_impl.go
```diff
@@ -0,0 +1,297 @@
...
+				continue
+			}
+		}
+
+		logger.Debugf("looking up cache key: %s", cacheKey.Key)
+		keysToGet = append(keysToGet, cacheKey.Key)
+	}
+
+	// Now fetch the pipeline.
+	responseDescriptorStatuses := make([]*pb.RateLimitResponse_DescriptorStatus,
```
nezdolik, line 110: Is `pipeline` term still applicable? (it is heavily used for redis, not aware how much for memcache)

src/memcached/cache_impl.go
```diff
@@ -0,0 +1,297 @@
...
+)
+
+type rateLimitMemcacheImpl struct {
+	client                     Client
+	timeSource                 utils.TimeSource
+	jitterRand                 *rand.Rand
+	expirationJitterMaxSeconds int64
+	cacheKeyGenerator          limiter.CacheKeyGenerator
+	localCache                 *freecache.Cache
+	wg                         sync.WaitGroup
```
nezdolik, line 48: nit: could expand field name to `waitGroup`

test/integration/integration_test.go
```diff
@@ -47,10 +47,10 @@ func newDescriptorStatusLegacy(
...
 //  stop the server at the end of each test, thus we can reuse the grpc port among these integration tests.
 func TestBasicConfig(t *testing.T) {
-	t.Run("WithoutPerSecondRedis", testBasicConfig("8083", "false", "0"))
-	t.Run("WithPerSecondRedis", testBasicConfig("8085", "true", "0"))
-	t.Run("WithoutPerSecondRedisWithLocalCache", testBasicConfig("18083", "false", "1000"))
-	t.Run("WithPerSecondRedisWithLocalCache", testBasicConfig("18085", "true", "1000"))
+	t.Run("WithoutPerSecondRedis", testBasicConfig("8083", "false", "0", ""))
+	t.Run("WithPerSecondRedis", testBasicConfig("8085", "true", "0", ""))
+	t.Run("WithoutPerSecondRedisWithLocalCache", testBasicConfig("18083", "false", "1000", ""))
+	t.Run("WithPerSecondRedisWithLocalCache", testBasicConfig("18085", "true", "1000", ""))
```
nezdolik, line 53: should there be also test for backend type "redis"?

test/memcached/cache_impl_test.go
```diff
@@ -0,0 +1,526 @@
...
+	stats "github.com/lyft/gostats"
+
+	"github.com/envoyproxy/ratelimit/test/common"
+	mock_memcached "github.com/envoyproxy/ratelimit/test/mocks/memcached"
+	mock_utils "github.com/envoyproxy/ratelimit/test/mocks/utils"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMemcached(t *testing.T) {
```
nezdolik, line 29: as a matter of preference i would have split that test in 3 smaller ones

test/memcached/cache_impl_test.go
```diff
@@ -0,0 +1,526 @@
...
+		cache.DoLimit(nil, request, limits))
+	assert.Equal(uint64(3), limits[0].Stats.TotalHits.Value())
+	assert.Equal(uint64(3), limits[0].Stats.OverLimit.Value())
+	assert.Equal(uint64(0), limits[0].Stats.NearLimit.Value())
+
+	cache.Flush()
+}
+
+func TestMemcacheWithJitter(t *testing.T) {
+	assert := assert.New(t)
```
nezdolik, line 413: every test sets up the same amount of routine code, this could be extracted to a sort of SetUp() function, to reduce the size of the test body and leave the essential part

test/memcached/cache_impl_test.go
nezdolik, line 29: would it be worth adding tests that simulate error responses from memcache backend?

nezdolik:
@dweitzman i could take a stab at removing code duplication. Lmk if this is fine with you or if you would prefer to do it yourself.

dweitzman:
> @dweitzman i could take a stab at removing code duplication. Lmk if this is fine with you or if you would prefer to do it yourself.

If it gets done and I don't have to do it, I'm perfectly happy :)

Sorry I've been a bit slow updating this PR for to address your comments. I'll try to get an update by Wednesday night.

nezdolik:
Thanks @dweitzman. I will be able to help out with duplicated code next week if this is fine.

test/integration/integration_test.go
dweitzman: "redis" is the default, but I modified TestBasicConfig so that some of the tests explicitly set "redis" instead of the empty string

test/memcached/cache_impl_test.go
dweitzman: This is intended to exactly mirror the test cases in the redis version. I wanted to keep it similar so there's hope of potentially extracting the test cases into some common location to share between redis and memcache, although I'm still not sure exactly what that looks like.

dweitzman: Added TestMemcachedGetError which tests both an explicit error from memcache and no explicit error but a response that doesn't contain the expected key

dweitzman (commented):
Addressed some of the review comments. The force push is because I realized immediately after pushing that my commit didn't have a "Signed-off-by", but not previous history before the new is modified

nezdolik:
@dweitzman started refactoring of duplicated code here: https://github.com/nezdolik/ratelimit/commit/1ff507a2aa38735a7bb9c3e0ae9ab05785cb0048 (it's based on latest state from your branch). Planning to finish it tomorrow and then we could add those commits to this PR.

src/memcached/cache_impl.go
```diff
@@ -0,0 +1,297 @@
...
+			limits[i].Stats.OverLimitWithLocalCache.Add(uint64(hitsAddend))
+			continue
+		}
+
+		rawMemcacheValue, ok := memcacheValues[cacheKey.Key]
+		var limitBeforeIncrease uint32
+		if ok {
+			decoded, err := strconv.ParseInt(string(rawMemcacheValue.Value), 10, 32)
+			if err != nil {
+				logger.Errorf("Unexpected non-numeric value in memcached: %v", rawMemcacheValue)
```
nezdolik, line 152: should there be `continue` keyword here? Does it make sense to go further and calculate limits with default initialised `int32`?

src/memcached/cache_impl.go
dweitzman: Changed

src/memcached/cache_impl.go
dweitzman: Seems like this case would still need the calculation. An extreme example would be if hitsAdded was high and the limit was low

Seems like someone might also set a quota of 0 if they wanted to stop traffic entirely

nezdolik: got it

nezdolik:
@dweitzman finished code refactoring, need to add unit tests for new code. I could also create a temp PR if it would be more convenient to review. Please take a look (it's my first Go code): https://github.com/nezdolik/ratelimit/commit/5842b4b16df3fe85b3d2979907659ed69bcb53c1

nezdolik:
it may be easier to merge existing PR and then refactoring PR on top of it. Created refactoring PR here: https://github.com/envoyproxy/ratelimit/pull/202, will rebase it on master once this PR is merged.

nezdolik (approved):

mattklein123 (approved):
SGTM to merge and then refactor. Thanks!

PR #173, opened 2020-09-10
Implement LOG_FORMAT=json

svdba:
Centralized log collection system works better with logs in json format.
E.g. DataDog strongly encourage setting up your logging library to produce your logs in JSON format to avoid the need for custom parsing rules.
So, the next small fix is all we need to get json logs.

Signed-off-by: Sergei Belyaev <sbelyaev@setronica.com>


mattklein123:
Looks good to me but please fix DCO and add documentation in the README. If there is any easy way to add a test that would be great also. Thank you!

svdba:
> Looks good to me but please fix DCO and add documentation in the README. If there is any easy way to add a test that would be great also. Thank you!
@mattklein123 The README was updated

mattklein123 (approved):
Thanks!

Issue #165, opened 2020-08-07
Support for ratelimit-reset


ioggstream:
## I expect

The `ratelimit` service to expose the data required to populate the `x-ratelimit-reset` header

## Instead

They are not exposed

## Reproduce

When I run 

```
git clone https://github.com/ioggstream/envoy-ratelimit-quickstart
cd envoy-ratelimit-quickstart
docker-compose up -d
curl -i http://localhost:8080/service/1
```

I get

```
HTTP/1.1 200 OK
content-type: text/html; charset=utf-8
content-length: 91
server: envoy
date: Fri, 07 Aug 2020 15:15:03 GMT
x-envoy-upstream-service-time: 11
x-ratelimit-limit: 50, 50;w=60
x-ratelimit-remaining: 49
x-ratelimit-reset: 0

Hello from behind Envoy (service 2)! hostname: 5eac71206ff9 resolvedhostname: 192.168.32.2
```
 
I expect

```
x-ratelimit-reset: 49
```

Instead

```
x-ratelimit-reset: 0
```

## Note

Thanks for implementing ratelimit draft in envoy :D

mattklein123:
I believe @Pchelolo is going to implement this.

Pchelolo:
Yup, just waiting for the new version of go-control-plane to be cut.

ioggstream:
Yes, but I thought an issue was worth opening :) Great work folks! :clap:

ioggstream:
Hi @Pchelolo, are there any news on the control plane release?

Pchelolo:
> Hi @Pchelolo, are there any news on the control plane release?

Hi, I'm not a maintainer, so I don't know. Maybe @mattklein123 can tell us when is a new release of the go-control-plane scheduled.

mattklein123:
I don't know what the release strategy of go-control-plane is. I think most people just import it by SHA. Can you inquire in that project?

ioggstream:
Hi @kyessenov, do you have any hints about the release date of the new go-control-plane?
Thanks for your support :)

kyessenov:
@ioggstream you can always use a particular SHA, `go get ..@SHA`. It's usually stable at head.

mattklein123:
> @ioggstream you can always use a particular SHA, go get ..@SHA. It's usually stable at head.

+1

ioggstream:
[...]

PR #182, opened 2020-10-20
Add support for x-ratelimit-reset header

clarakosi:
Signed-off-by: Clara Andrew-Wani <candrewwani@gmail.com>

Issue: #165

### DIFF
--- a/src/redis/cache_impl.go
+++ b/src/redis/cache_impl.go
@@ -214,13 +214,13 @@ func (this *rateLimitCacheImpl) DoLimit(
 	return responseDescriptorStatuses
 }
 
-func CalculateReset(currentLimit *pb.RateLimitResponse_RateLimit ) *duration.Duration {
+func CalculateReset(currentLimit *pb.RateLimitResponse_RateLimit) *duration.Duration {
 	sec := unitInSeconds(currentLimit.Unit)
 	now := limiter.NewTimeSourceImpl().UnixNow()
 	return &duration.Duration{Seconds: sec - now%sec}
 }
 
-func unitInSeconds( unit pb.RateLimitResponse_RateLimit_Unit) int64 {
+func unitInSeconds(unit pb.RateLimitResponse_RateLimit_Unit) int64 {
 	switch unit {
 	case pb.RateLimitResponse_RateLimit_SECOND:
 		return 1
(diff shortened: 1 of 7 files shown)


src/redis/cache_impl.go
```diff
@@ -210,6 +214,27 @@ func (this *rateLimitCacheImpl) DoLimit(
 	return responseDescriptorStatuses
 }
 
+func CalculateReset(currentLimit *pb.RateLimitResponse_RateLimit) *duration.Duration {
+	sec := unitInSeconds(currentLimit.Unit)
+	now := limiter.NewTimeSourceImpl().UnixNow()
+	return &duration.Duration{Seconds: sec - now%sec}
+}
+
+func unitInSeconds(unit pb.RateLimitResponse_RateLimit_Unit) int64 {
```
mattklein123, line 223: We already have this function here: https://github.com/envoyproxy/ratelimit/blob/e91321b10f1ad7691d0348e880bd75d0fca05758/src/limiter/cache_key.go#L39

Can you factor it into a utility?

src/redis/cache_impl.go
```diff
@@ -210,6 +214,27 @@ func (this *rateLimitCacheImpl) DoLimit(
 	return responseDescriptorStatuses
 }
 
+func CalculateReset(currentLimit *pb.RateLimitResponse_RateLimit) *duration.Duration {
+	sec := unitInSeconds(currentLimit.Unit)
+	now := limiter.NewTimeSourceImpl().UnixNow()
```
mattklein123, line 219: Does creating a time source cost anything perf-wise? Can you pre-create the time source and/or pass it in to the constructor?

test/integration/integration_test.go
```diff
@@ -3,19 +3,17 @@
...
 	"math/rand"
-	"net/http"
 	"os"
 	"strconv"
 	"testing"
 	"time"
 
 	pb_legacy "github.com/envoyproxy/go-control-plane/envoy/service/ratelimit/v2"
 	pb "github.com/envoyproxy/go-control-plane/envoy/service/ratelimit/v3"
+	"github.com/envoyproxy/ratelimit/src/redis"
```
mattklein123, line 16: I would prefer that you not use the utility here to compute the limit. I think the tests would be stronger if you hand computed them and then we can check to see if they match?

mattklein123 (changes requested):
Thank you! Just a few small comments.

mattklein123 (approved):
Thanks!

PR #187, opened 2020-11-03
Create repokitteh.star

itayd:
Initially only support `/wait`. 

Signed-off-by: Itay Donanhirsh itay@bazoo.org


itayd:
bla bla to unwait?

itayd:
bla bla to unwait

PR #192, opened 2020-11-17
Fix flakey tests with DurationUntilReset. Update docker example to V3 config.

ysawa0:
- Updated existing integration tests such that DurationUntilReset is not tested (guaranteed to be the same) due to flakiness
- Added new test for DurationUntilReset
- Update docker compose example to V3 API.
- Add example prom statsd exporter config.


test/integration/integration_test.go
```diff
@@ -467,6 +466,21 @@ func testBasicBaseConfig(grpcPort, perSecond string, local_cache_size string) fu
...
+			context.Background(),
+			common.NewRateLimitRequest("another", [][][2]string{{{getCacheKey("key4", enable_local_cache), "durTest"}}}, 1))
+
+		time.Sleep(2 * time.Second) // Wait to allow duration to tick down
+
+		resp2, err := c.ShouldRateLimit(
+			context.Background(),
+			common.NewRateLimitRequest("another", [][][2]string{{{getCacheKey("key4", enable_local_cache), "durTest"}}}, 1))
+
+		if resp2.GetStatuses()[0].DurationUntilReset != nil {
```
mattklein123, line 481: Shouldn't we actually test that this isn't nil? When would it be nil?

mattklein123 (changes requested):
Thanks a ton for fixing this. One small comment.

/wait

test/integration/integration_test.go
ysawa0: I've been looking into this and there seems to be a bug where if local cache is on and a valid request like
`domain:"foo"  descriptors:{entries:{key:"hello"  value:"world"}}  hits_addend:1` is sent, the code thinks the key is empty and therefore returns a `overall_code:OK  statuses:{code:OK}` response with no duration until reset. Still digging into this, stay tuned...

ysawa0: Okay, fixed. Was just missing the corresponding local cache key in configs.

mattklein123 (approved):
Thanks!

PR #194, opened 2020-11-24
Separate Redis cache and driver implementation

walbertus:
Separate Redis cache and driver implementation into three parts:
1. Fixed window rate limit cache implementation
2. Test for Redis driver implementation
3. Constructor of rate limit cache implementation


mattklein123 (approved):
Awesome, thanks for splitting this out.

PR #197, opened 2020-11-26
Add debug logging to indicate descriptor and limit

AlexKulbiy:
We are actively using ratelimit service with istio. With multiple levels of abstraction (Istio EnvoyFilter -> Envoy configuration -> ratelimit descriptor) it's sometimes hard to understand how does the descriptor received by the specific request look like.

This PR adds three debug level log messages, which should help understanding what exactly is used by the rate limit at moment.

Example messages:

simple descriptor:
```
time="2020-11-26T16:10:43Z" level=debug msg="got descriptor: (header_match=path)"
```

compound descriptor:
```
time="2020-11-26T17:10:22Z" level=debug msg="got descriptor: (header_match=get),(generic_key=foo)"
```

rate limit applied:
```
time="2020-11-26T16:10:43Z" level=debug msg="applying limit: 20 requests per SECOND"
```

no matches:
```
time="2020-11-26T16:22:27Z" level=debug msg="descriptor does not match any limit, no limits applied"
```

### DIFF
--- a/src/service/ratelimit.go
+++ b/src/service/ratelimit.go
@@ -114,23 +114,27 @@ func (this *service) shouldRateLimitWorker(
 
 	limitsToCheck := make([]*config.RateLimit, len(request.Descriptors))
 	for i, descriptor := range request.Descriptors {
-		var descriptorEntryStrings []string
-		for _, descriptorEntry := range descriptor.GetEntries() {
-			descriptorEntryStrings = append(
-				descriptorEntryStrings,
-				fmt.Sprintf("(%s=%s)", descriptorEntry.Key, descriptorEntry.Value),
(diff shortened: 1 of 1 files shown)


mattklein123 (changes requested):
Thanks this makes sense to me, but is it possible to refactor slightly so we don't pay the compute cost if the log level is not debug? I think it shouldn't be too hard. Thank you!

/wait

AlexKulbiy:
Thanks for the feedback, I didn't think about this and it makes total sense. I've made a change to verify log level before calculating values for debug logging. Please let me know how does that look.

mattklein123 (approved):
Thanks!

Issue #199, opened 2020-12-02
envconfig cannot convert REDIS_PIPELINE_WINDOW to time duration


zufardhiyaulhaq:
we find this error when configuring `REDIS_PIPELINE_WINDOW`:
```
panic: envconfig.Process: assigning _REDIS_PIPELINE_WINDOW to REDIS_PIPELINE_WINDOW: converting '1' to type time.Duration

goroutine 1 [running]:
github.com/envoyproxy/ratelimit/src/settings.NewSettings(...)
	/ratelimit/src/settings/settings.go:55
github.com/envoyproxy/ratelimit/src/service_cmd/runner.(*Runner).Run(0xc000197f68)
	/ratelimit/src/service_cmd/runner/runner.go:39 +0x973
main.main()
	/ratelimit/src/service_cmd/main.go:7 +0x43
```

we have configmap and added the configmap as a variable in Kubernetes:
```
apiVersion: v1
data:
  LOCAL_CACHE_SIZE_IN_BYTES: "32767"
  LOG_LEVEL: info
  REDIS_PIPELINE_LIMIT: "1"
  REDIS_PIPELINE_WINDOW: "1"
  REDIS_POOL_SIZE: "40"
  REDIS_TYPE: cluster
  REDIS_URL: redis-cluster-1:6379, redis-cluster-2:6379, redis-cluster-3:6379
  STATSD_HOST: statsd-exporter.infrastructure.svc
  USE_STATSD: "true"
kind: ConfigMap
metadata:
  name: rate-limit-env
---
    envFrom:
    - configMapRef:
        name: rate-limit-env
```

caitong93:
I think the issue here is `REDIS_PIPELINE_WINDOW` expect a time duration but receive an integer,  due the lack of config validation(and also lack of explanation).
There is one advantage to use time duration, it allows for more granularity, e.g. `100us`, `1ms`(radix use 150us by default). 
How about keeping `REDIS_PIPELINE_WINDOW` as time duration and add config validation instead? Does that sound reasonable to you?

zufardhiyaulhaq:
@caitong93 yes this is already resolved, we are not aware that its use time duration from the README.md, probably update documentation or create a config validation.

PR #200, opened 2020-12-02
Fix error when set REDIS_PIPELINE_WINDOW and REDIS_PERSECOND_PIPELINE_WINDOW

madebyais:
According to https://github.com/envoyproxy/ratelimit/issues/199 issue, we fix it by adding new field in the settings struct and adding additional function to convert from int64 into time.Duration.

Co-authored-by: zufardhiyaulhaq <zufardhiyaulhaq@gmail.com>

### DIFF
--- a/src/redis/cache_impl.go
+++ b/src/redis/cache_impl.go
@@ -2,6 +2,7 @@ package redis
 
 import (
 	"math/rand"
+	"time"
 
 	"github.com/coocood/freecache"
 	"github.com/envoyproxy/ratelimit/src/limiter"
@@ -10,14 +11,17 @@ import (
 )
 
 func NewRateLimiterCacheImplFromSettings(s settings.Settings, localCache *freecache.Cache, srv server.Server, timeSource limiter.TimeSource, jitterRand *rand.Rand, expirationJitterMaxSeconds int64) limiter.RateLimitCache {
+	redisPipelineWindow := time.Duration(s.RedisPipelineWindow) * time.Second
+
 	var perSecondPool Client
 	if s.RedisPerSecond {
 		perSecondPool = NewClientImpl(srv.Scope().Scope("redis_per_second_pool"), s.RedisPerSecondTls, s.RedisPerSecondAuth,
-			s.RedisPerSecondType, s.RedisPerSecondUrl, s.RedisPerSecondPoolSize, s.RedisPipelineWindow, s.RedisPipelineLimit)
+			s.RedisPerSecondType, s.RedisPerSecondUrl, s.RedisPerSecondPoolSize, redisPipelineWindow, s.RedisPipelineLimit)
 	}
+
 	var otherPool Client
 	otherPool = NewClientImpl(srv.Scope().Scope("redis_pool"), s.RedisTls, s.RedisAuth, s.RedisType, s.RedisUrl, s.RedisPoolSize,
-		s.RedisPipelineWindow, s.RedisPipelineLimit)
+		redisPipelineWindow, s.RedisPipelineLimit)
 
 	return NewFixedRateLimitCacheImpl(
 		otherPool,
--- a/src/settings/settings.go
+++ b/src/settings/settings.go
@@ -1,8 +1,6 @@
 package settings
 
 import (
-	"time"
-
 	"github.com/kelseyhightower/envconfig"
 	"google.golang.org/grpc"
 )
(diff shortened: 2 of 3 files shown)


zufardhiyaulhaq:
we already test this and build the docker image if you need to check it directly
https://hub.docker.com/r/zufardhiyaulhaq/ratelimit/

src/settings/settings.go
```diff
@@ -11,38 +11,45 @@ type Settings struct {
...
+	RedisPerSecondPipelineLimit         int     `envconfig:"REDIS_PERSECOND_PIPELINE_LIMIT" default:"0"`
+	ExpirationJitterMaxSeconds          int64   `envconfig:"EXPIRATION_JITTER_MAX_SECONDS" default:"300"`
+	LocalCacheSizeInBytes               int     `envconfig:"LOCAL_CACHE_SIZE_IN_BYTES" default:"0"`
+	NearLimitRatio                      float32 `envconfig:"NEAR_LIMIT_RATIO" default:"0.8"`
+}
+
+func (s *Settings) ConvertRedisPipelineValueToTimeDuration() {
+	s.RedisPipelineWindow = time.Duration(s.RedisPipelineWindowInInt64) * time.Second
+	s.RedisPerSecondPipelineWindow = time.Duration(s.RedisPerSecondPipelineWindowInInt64) * time.Second
 }
```
mattklein123, line 53: Instead of a new config field, can you just try to convert from each type? I think that would be OK? Also I'm not sure how this couldn't wouldn't overwrite the other settings if set.

test/settings/settings_test.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/envoyproxy/ratelimit/src/settings"
+)
+
+func TestShouldReturnDefaultValueForRedisPipelineWindow(t *testing.T) {
+	expectedValue := time.Duration(0) * time.Second
+
+	s := settings.NewSettings()
+	if s.RedisPipelineWindow != expectedValue {
+		t.Errorf("Got %s, expected %s", s.RedisPipelineWindow, expectedValue)
+	}
```
mattklein123, line 17: Please use testing assertions like the other tests.

mattklein123 (changes requested):
Thanks for fixing. A few small comments.

/wait

zufardhiyaulhaq:
@mattklein123 we cannot find any implementation for `RedisPerSecondPipelineWindow`.

test/settings/settings_test.go
madebyais: Hi @mattklein123 I've updated the test for the settings package to use the test suite by github.com/stretchr/testify.

Please kindly review. Thank you.

src/settings/settings.go
madebyais: I did another changes, by removing this function and replacing the type for both `RedisPipelineWindow` and `RedisPerSecondPipelineWindow` to be `int64` instead of `time.Duration`. Then do the conversion in the `NewRateLimiterCacheImplFromSettings` in the `redis/cache_impl.go`

Kindly let me know if you are good with this approach.

src/redis/cache_impl.go
```diff
@@ -10,14 +11,17 @@ import (
 )
 
 func NewRateLimiterCacheImplFromSettings(s settings.Settings, localCache *freecache.Cache, srv server.Server, timeSource limiter.TimeSource, jitterRand *rand.Rand, expirationJitterMaxSeconds int64) limiter.RateLimitCache {
+	redisPipelineWindow := time.Duration(s.RedisPipelineWindow) * time.Second
+
 	var perSecondPool Client
 	if s.RedisPerSecond {
 		perSecondPool = NewClientImpl(srv.Scope().Scope("redis_per_second_pool"), s.RedisPerSecondTls, s.RedisPerSecondAuth,
-			s.RedisPerSecondType, s.RedisPerSecondUrl, s.RedisPerSecondPoolSize, s.RedisPipelineWindow, s.RedisPipelineLimit)
+			s.RedisPerSecondType, s.RedisPerSecondUrl, s.RedisPerSecondPoolSize, redisPipelineWindow, s.RedisPipelineLimit)
```
mattklein123, line 19: Yeah I think this was a pre-existing bug. Can you fix this please?

Also, can you explain the underlying issue in more detail? Is this change from time to int going to break any existing users? Does this need any doc update? Thank you.

/wait

ysawa0: Also, correct me if I'm wrong but I noticed the perSecondPool client should be using the value from `s.RedisPerSecondPipelineWindow` and `s.RedisPerSecondPipelineLimit` 
Right now it looks like its using the same value as the regular, non-seconds client.

zufardhiyaulhaq:
Hi, we already sort this issue https://github.com/envoyproxy/ratelimit/issues/199 and this PR can probably closed. Adding validation as @caitong93 suggested is more reasonable, or perhaps update the README with better documentation on how to fill `REDIS_PIPELINE_WINDOW` since not all users can read the code. @madebyais can you close this issue, sir?

PR #210, opened 2021-01-28
Add CACHE_KEY_PREFIX for an optional redis/memcache key prefix

dweitzman:
k/v store proxies like envoy with redis protocol or mcrouter with memcached
protocol often support prefix-specific routing rules to direct different types
of cache traffic to different backend clusters.


mattklein123 (approved):
Thanks!

PR #211, opened 2021-01-28
Report memcached client hit/miss/error stats to allow monitoring for memcache outages

dweitzman:
The rate limiter itself fails open and swallows errors, so reporting k/v store client statistics
is helpful for catching a backend outage.

(I've chosen not to include latency statistics here because statsd doesn't have an official protocol
and I've found through experimentation that the lyft stats library output format for timing stats
isn't compatible with our metric collector)

### DIFF
--- a/src/memcached/stats_collecting_client.go
+++ b/src/memcached/stats_collecting_client.go
@@ -20,7 +20,7 @@ type statsCollectingClient struct {
 	keysFound        stats.Counter
 }
 
-func collectStats(c Client, scope stats.Scope) Client {
+func CollectStats(c Client, scope stats.Scope) Client {
 	return statsCollectingClient{
 		c:                c,
 		multiGetSuccess:  scope.NewCounterWithTags("multiget", map[string]string{"code": "success"}),
--- a/src/memcached/stats_collecting_client.go
+++ b/src/memcached/stats_collecting_client.go
@@ -1,0 +1,80 @@
+package memcached
+
+import (
+	"github.com/bradfitz/gomemcache/memcache"
+	stats "github.com/lyft/gostats"
+)
+
+type statsCollectingClient struct {
+	c Client
+
+	multiGetSuccess  stats.Counter
+	multiGetError    stats.Counter
+	incrementSuccess stats.Counter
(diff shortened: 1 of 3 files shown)


mattklein123 (changes requested):
Very nice. Is it possible to add some tests for these stats?

/wait

dweitzman:
Added unit tests

dweitzman:
> Very nice. Is it possible to add some tests for these stats?
> 
> /wait

Done

src/memcached/stats_collecting_client.go
```diff
@@ -0,0 +1,80 @@
...
+		multiGetSuccess:  scope.NewCounterWithTags("multiget", map[string]string{"code": "success"}),
+		multiGetError:    scope.NewCounterWithTags("multiget", map[string]string{"code": "error"}),
+		incrementSuccess: scope.NewCounterWithTags("increment", map[string]string{"code": "success"}),
+		incrementMiss:    scope.NewCounterWithTags("increment", map[string]string{"code": "miss"}),
+		incrementError:   scope.NewCounterWithTags("increment", map[string]string{"code": "error"}),
+		addSuccess:       scope.NewCounterWithTags("add", map[string]string{"code": "success"}),
+		addError:         scope.NewCounterWithTags("add", map[string]string{"code": "error"}),
+		addNotStored:     scope.NewCounterWithTags("add", map[string]string{"code": "not_stored"}),
+		keysRequested:    scope.NewCounter("keys_requested"),
+		keysFound:        scope.NewCounter("keys_found"),
```
mattklein123, line 35: I don't see any tests for these? Can you make sure they are all covered? Thank you.

/wait

dweitzman: keys_requested and keys_found are tested during the tests for GetMulti. I've added one more test just now so there are tests where keys_found and keys_requested are either 1 or 2.

mattklein123: Sorry, missed this. Awesome, thanks!

mattklein123 (approved):

PR #212, opened 2021-01-29
Begin changing integration tests to explicitly run redis/memcache as a subprocess

dweitzman:
The motivation for this is that the "tests_with_redis" rule in the Makefile is a giant
unwieldy mess and it's hard to know which redis/memcache instances are used by which tests.
The tests themselves hardcode mysterious port numbers whose meaning you could only know by
cross-referencing the makefile.

Some benefits to running the cache processes on-demand:

- The tests themselves become more isolated, complete, and clear
- "go test" could work out of the box for many people, no docker
  or pre-configured cache services on special ports necessary.
  That seems like it'd be a big win for ease of testing and development of
  integration tests.
- A next step after making "go test" work out of the box could be
  dynamic free port selection so that if someone happens to already
  have a service running on a particular port it wouldn't block "go test"
  from working.

Another small change included here is to make the memcache integration tests
use synchronous rather than async increments to avoid a race condition checking
whether quota is being depleted.

One last change is ensure that the test name is correctly printed when
a panic() happens within the rate limit service in an integration test.
The slight tradeoff is that the original stack trace is lost, but a
stack trace without a test name in it isn't particularly helpful for
debugging.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,12 +73,8 @@ tests: compile
 
 .PHONY: tests_with_redis
 tests_with_redis: bootstrap_redis_tls tests_unit
-	redis-server --port 6379 &
-	redis-server --port 6380 &
 	redis-server --port 6381 --requirepass password123 &
 	redis-server --port 6382 --requirepass password123 &
-	redis-server --port 6384 --requirepass password123 &
-	redis-server --port 6385 --requirepass password123 &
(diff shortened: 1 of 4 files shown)


mattklein123 (approved):
Awesome, thanks!

PR #213, opened 2021-01-31
Resolve uncaught merge conflict

dweitzman:
I had two PRs out at the same time, one adding an integration test and one refactoring how some integration tests work. This resolves a conflict

### DIFF
--- a/src/server/server.go
+++ b/src/server/server.go
@@ -39,5 +39,8 @@ type Server interface {
 	 */
 	Runtime() loader.IFace
 
+	/**
+	 *  Stops serving the grpc port (for integration testing).
+	 */
 	Stop()
 }
--- a/src/server/server.go
+++ b/src/server/server.go
@@ -1,11 +1,12 @@
 package server
 
 import (
-	pb "github.com/envoyproxy/go-control-plane/envoy/service/ratelimit/v3"
 	"net/http"
 
+	pb "github.com/envoyproxy/go-control-plane/envoy/service/ratelimit/v3"
+
 	"github.com/lyft/goruntime/loader"
-	"github.com/lyft/gostats"
+	stats "github.com/lyft/gostats"
 	"google.golang.org/grpc"
 )
 
@@ -37,4 +38,6 @@ type Server interface {
 	 * Returns the runtime configuration for the server.
 	 */
 	Runtime() loader.IFace
+
+	Stop()
 }
--- a/src/server/server_impl.go
+++ b/src/server/server_impl.go
@@ -9,6 +9,7 @@ import (
 	"net/http/pprof"
 	"path/filepath"
 	"sort"
+	"sync"
 
 	"os"
 	"os/signal"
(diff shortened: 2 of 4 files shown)


mattklein123:
Check CI? Looks like it's still failing.

/wait

dweitzman:
I'm a little confused because CI says it checked out commit "760d9adb89a99d2cc8473dc4a5695afcda963d36", which is not either of the commits I have here. I'm wondering if CI might not really have tested the latest version?...

mattklein123:
I think it does a merge into main branch but I'm not 100% sure. I just fixed the main CI but I don't think that should impact PRs.

dweitzman:
I think something may actually be flaky. I can investigate in an hour or two

dweitzman:
I think I might have reproduced the issue one time, but I'm having trouble reproducing it again. I've run the *Prefix* tests a number of times in a row and they keep passing for me.

dweitzman:
Turns out the "CachePrefix" tests were just flat-out testing the wrong thing, going back to the opaqueness of mysterious port numbers.

I accidentally copy-pasted grpc port numbers from other tests. Since all grpcs Runner instances started during integration tests run forever (Runner doesn't even have a Stop method) the net effect was that these tests were pointing at wrong, old grpc ports

I'm testing an updated version of this PR that fixes the port numbers and also introduces a "Stop()" method to Runner to protect against similar mistakes in the future. Seems promising so far since I was able to reproduce issues when running the tests multiple times in a row with the "-count" option to go test and with the change I'm working on the tests are passing with "-count"

dweitzman:
It's also worth noting that Runner.Start() has no return error code, so if you try to start on a port that's already taken it'll silently fail without causing any related integration tests to fail

src/server/server.go
```diff
@@ -37,4 +38,6 @@ type Server interface {
 	 * Returns the runtime configuration for the server.
 	 */
 	Runtime() loader.IFace
+
+	Stop()
```
mattklein123, line 42: nit: please add method doc comment

src/server/server_impl.go
```diff
@@ -38,19 +39,22 @@ type serverDebugListener struct {
...
+	grpcPort        int
+	debugPort       int
+	router          *mux.Router
+	grpcServer      *grpc.Server
+	store           stats.Store
+	scope           stats.Scope
+	runtime         loader.IFace
+	debugListener   serverDebugListener
+	health          *HealthChecker
+	debugListenerMu sync.Mutex
```
mattklein123, line 52: nit: put this next to the listener

src/service_cmd/runner/runner.go
```diff
@@ -85,6 +88,9 @@ func (runner *Runner) Run() {
 	}
 
 	srv := server.NewServer("ratelimit", runner.statsStore, localCache, settings.GrpcUnaryInterceptor(nil))
+	runner.mu.Lock()
```
mattklein123, line 91: Why is this lock necessary? It's not immediately clear to me.

mattklein123 (changes requested):
Thanks for fixing. A few comments/questions.

/wait

src/service_cmd/runner/runner.go
dweitzman: Run() and Stop() are called from different threads, since Run() basically hangs forever when it's called in the test.

Since `srv` isn't initialized until Run() is called, seemed like a lock was needed here.

I think I added this lock just because it seemed to make sense for transferring a field between two goroutines. The other lock in server_impl was added because the race detector was failing the test.

src/server/server_impl.go
dweitzman: Done

src/server/server.go
dweitzman: Done

dweitzman:
Addressed comments

src/service_cmd/runner/runner.go
mattklein123: OK sounds good, thanks.

mattklein123 (approved):
Thanks!

PR #217, opened 2021-02-02
Improve logging + error handling

deepthawtz:
Some small logging changes we found helpful in running this service at Venmo.

### DIFF
--- a/src/server/server_impl.go
+++ b/src/server/server_impl.go
@@ -125,7 +125,7 @@ func (server *server) Start() {
 		server.debugListener.listener = lis
 		server.listenerMu.Unlock()
 
-		if err := http.Serve(server.debugListener.listener, server.debugListener.debugMux); err != nil {
+		if err := http.Serve(server.debugListener.listener, server.debugListener.debugMux); err != http.ErrServerClosed {
 			logger.Fatalf("Failed to start debug server: '%+v'", err)
 		}
 	}()
@@ -136,7 +136,7 @@ func (server *server) Start() {
 
 	addr := fmt.Sprintf(":%d", server.port)
 	logger.Warnf("Listening for HTTP on '%s'", addr)
-	list, err := reuseport.Listen("tcp", addr)
+	lis, err := reuseport.Listen("tcp", addr)
 	if err != nil {
 		logger.Fatalf("Failed to open HTTP listener: '%+v'", err)
 	}
@@ -144,10 +144,9 @@ func (server *server) Start() {
 	server.listenerMu.Lock()
 	server.httpServer = srv
 	server.listenerMu.Unlock()
-	err = srv.Serve(list)
 
-	if err != http.ErrServerClosed {
-		logger.Fatal(err)
+	if err := srv.Serve(lis); err != http.ErrServerClosed {
+		logger.Fatalf("Failed to start HTTP server: '%+v'", err)
 	}
 }
 
--- a/src/server/server_impl.go
+++ b/src/server/server_impl.go
@@ -126,7 +126,7 @@ func (server *server) Start() {
 			return
 		}
 		err = http.Serve(server.debugListener.listener, server.debugListener.debugMux)
-		logger.Infof("Failed to start debug server '%+v'", err)
+		logger.Errorf("Failed to start debug server '%+v'", err)
 	}()
 
 	go server.startGrpc()
--- a/src/server/server_impl.go
+++ b/src/server/server_impl.go
@@ -116,17 +116,18 @@ func (server *server) Start() {
 	go func() {
 		addr := fmt.Sprintf(":%d", server.debugPort)
 		logger.Warnf("Listening for debug on '%s'", addr)
-		var err error
-		server.listenerMu.Lock()
-		server.debugListener.listener, err = reuseport.Listen("tcp", addr)
-		server.listenerMu.Unlock()
-
+		lis, err := reuseport.Listen("tcp", addr)
 		if err != nil {
 			logger.Errorf("Failed to open debug HTTP listener: '%+v'", err)
 			return
 		}
-		err = http.Serve(server.debugListener.listener, server.debugListener.debugMux)
-		logger.Errorf("Failed to start debug server '%+v'", err)
+		server.listenerMu.Lock()
+		server.debugListener.listener = lis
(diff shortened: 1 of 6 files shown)


src/server/server_impl.go
```diff
@@ -125,7 +125,7 @@ func (server *server) Start() {
 			return
 		}
 		err = http.Serve(server.debugListener.listener, server.debugListener.debugMux)
-		logger.Infof("Failed to start debug server '%+v'", err)
+		logger.Errorf("Failed to start debug server '%+v'", err)
```
mattklein123, line 128: Shouldn't you check err above before logging this error?

src/service/ratelimit.go
```diff
@@ -225,7 +225,7 @@ func NewService(runtime loader.IFace, cache limiter.RateLimitCache,
 		for {
 			logger.Debugf("waiting for runtime update")
 			<-newService.runtimeUpdateEvent
-			logger.Debugf("got runtime update and reloading config")
+			logger.Warnf("got runtime update and reloading config")
```
mattklein123, line 228: I think info would be better?

src/service_cmd/runner/runner.go
```diff
@@ -84,6 +84,7 @@ func (runner *Runner) Run() {
 
 	var localCache *freecache.Cache
 	if s.LocalCacheSizeInBytes != 0 {
+		logger.Warnf("initializing local cache with %d bytes", s.LocalCacheSizeInBytes)
```
mattklein123, line 87: Info seems better?

mattklein123 (changes requested):
Thanks for the improvements. Some small comments.

/wait

src/server/server_impl.go
deepthawtz: good point! Code was already written like that w/o checking err. Might as well fix while I'm at it.

src/service/ratelimit.go
deepthawtz: Info would work too, used Warn level because that was the default log level in settings.go and there are almost no Info level logs in the code base.

src/server/server_impl.go
mattklein123: This was fixed in https://github.com/envoyproxy/ratelimit/pull/219. Can you merge main and then I think you can revert this part?

src/service/ratelimit.go
mattklein123: Yeah I think that is historical, though I think it's a mistake (of my own making). WDYT about changing the default to info and changing the various warns that should be info to info?

mattklein123 (changes requested):

src/service/ratelimit.go
deepthawtz: I like that. Would feel much more natural that way.

src/server/server_impl.go
deepthawtz: I rebased those changes just now.  #219 was a different call to `Serve()` so this code still exists.

In #219 I'm not sure about the need to specifically check the type of error is `http.ErrServerClosed` though, why not simply handle all non-nil errors so we anticipate all possible errors?

src/server/server_impl.go
```diff
@@ -115,49 +115,51 @@ func (server *server) GrpcServer() *grpc.Server {
...
 	srv := &http.Server{Handler: server.router}
 	server.listenerMu.Lock()
 	server.httpServer = srv
 	server.listenerMu.Unlock()
-	err = srv.Serve(list)
 
-	if err != http.ErrServerClosed {
-		logger.Fatal(err)
+	if err := srv.Serve(lis); err != nil {
+		logger.Fatalf("Failed to start HTTP server: '%+v'", err)
```
mattklein123, line 149: There are cases in which we do a graceful shutdown and we don't want to print the fatal and exit. This is actually the exit case, so I think you want the code that was there before, and I think you want to replicate this above and below also? cc @dweitzman

mattklein123 (changes requested):
Thanks, small comment.

/wait

src/server/server_impl.go
deepthawtz: 💡 I get it now, makes perfect sense. I can correct that.

mattklein123:
CI appears to be hanging for some reason. Can you take a look? Otherwise LGTM.

/wait

src/server/server_impl.go
dweitzman: Looks good. Thanks

dweitzman:
In case this is helpful with debugging: I've noticed that poorly-timed panic()s from with goroutines or possibly the race detector can cause the integration tests to just hang. Even `go test -v` can fail to help because it really likes to print test names first and it can't always do that if there's a poorly-timed panic.

I find that modifying the `Makefile` command in `tests_with_redis` to something like `cd test/integration && go test -v -race -tags=integration` will print the actual error that led to the tests getting stuck. The `cd` in that command actually does matter, because then go doesn't need to worry about the test output format and wait to know the test name.

This might all be more reliable if more of the goroutines caught panic()s and communicated the failures back to their parent goroutines.

dweitzman:
I also wonder if using `grpc.DialContext()` instead of `grpc.Dial()` in the integration tests might help avoid hanging.

I'm not sure exactly why hangs happen (haven't dug into it), but I can imagine it happening because the listener accepts a TCP connection but the server is too dead to actually respond to it.

dweitzman:
Ah, I think I understand the hanging now. It's related to the new subprocesses I added for memcache/redis somehow keeping the parent process alive after a panic(), but it also only happens because of the subprocess is sharing stdout and stderr with the parent process.

For quick debugging you can change `startCacheProcess` in common.go like this, but it'll only compile on linux:
```
	cmd := exec.CommandContext(ctx, command, args...)
	cmd.SysProcAttr = &syscall.SysProcAttr{
		Pdeathsig: syscall.SIGTERM,
	}
```

I'm going to modify `startCacheProcess` to pipe stdout and stderr through memory in go instead of sharing file descriptors

dweitzman:
Mailed https://github.com/envoyproxy/ratelimit/pull/221

deepthawtz:
Thanks @mattklein123 and @dweitzman. I'm going to close this temporarily until I can sort this out

Issue #119, opened 2020-01-10
Add Stopping the server into the runner


freedomljc:
We'll need to be able to stop the server in the runner, thus cleaning up after the integration test.

PR #219, opened 2021-02-03
integration test refactoring: better isolation, typed config, and fewer magic numbers

dweitzman:
1) All tests use the same grpc port now that the Runner.Stop() exists, removing some magic numbers and surface area for mistakes with those numbers

2) Rather than doing os.Setenv(), tests construct a typed new Settings object for each run.

The downside here is that the settings library itself won't get exercised in terms of how it parses env variables.

The upsides are:
- Static type checking gives protection against typos in integration test setup and IDE code completion
- Eliminates the risk of accidentally leaving env variables unchanged between tests, ensuring
  every test starts with a clean slate.

This resolves https://github.com/envoyproxy/ratelimit/issues/119

### DIFF
--- a/src/server/server_impl.go
+++ b/src/server/server_impl.go
@@ -39,22 +39,24 @@ type serverDebugListener struct {
 }
 
 type server struct {
-	port            int
-	grpcPort        int
-	debugPort       int
-	router          *mux.Router
-	grpcServer      *grpc.Server
-	store           stats.Store
-	scope           stats.Scope
-	runtime         loader.IFace
-	debugListener   serverDebugListener
-	debugListenerMu sync.Mutex
-	health          *HealthChecker
+	port          int
+	grpcPort      int
+	debugPort     int
+	router        *mux.Router
+	grpcServer    *grpc.Server
+	store         stats.Store
+	scope         stats.Scope
+	runtime       loader.IFace
+	debugListener serverDebugListener
+	httpListener  net.Listener
+	listenerMu    sync.Mutex
+	health        *HealthChecker
+	stopped       bool
 }
 
 func (server *server) AddDebugHttpEndpoint(path string, help string, handler http.HandlerFunc) {
-	server.debugListenerMu.Lock()
-	defer server.debugListenerMu.Unlock()
+	server.listenerMu.Lock()
+	defer server.listenerMu.Unlock()
 	server.debugListener.debugMux.HandleFunc(path, handler)
 	server.debugListener.endpoints[path] = help
 }
@@ -116,9 +118,9 @@ func (server *server) Start() {
 		addr := fmt.Sprintf(":%d", server.debugPort)
 		logger.Warnf("Listening for debug on '%s'", addr)
 		var err error
-		server.debugListenerMu.Lock()
+		server.listenerMu.Lock()
 		server.debugListener.listener, err = reuseport.Listen("tcp", addr)
-		server.debugListenerMu.Unlock()
+		server.listenerMu.Unlock()
 
 		if err != nil {
 			logger.Errorf("Failed to open debug HTTP listener: '%+v'", err)
@@ -138,7 +140,20 @@ func (server *server) Start() {
 	if err != nil {
 		logger.Fatalf("Failed to open HTTP listener: '%+v'", err)
 	}
-	logger.Fatal(http.Serve(list, server.router))
+	server.listenerMu.Lock()
+	server.httpListener = list
+	server.listenerMu.Unlock()
+	err = http.Serve(list, server.router)
+
+	server.listenerMu.Lock()
+	explicitlyStopped := server.stopped
+	server.listenerMu.Unlock()
+
+	if !explicitlyStopped {
+		logger.Fatal(err)
+	} else {
+		logger.Info(err)
+	}
 }
 
 func (server *server) startGrpc() {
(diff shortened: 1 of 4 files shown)


test/integration/integration_test.go
```diff
@@ -45,239 +75,230 @@ func newDescriptorStatusLegacy(
...
+	return s
+}
+
 func TestBasicConfig(t *testing.T) {
 	common.WithMultiRedis(t, []common.RedisConfig{
-		{Port: 6379},
+		{Port: 6383},
 		{Port: 6380},
 	}, func() {
-		t.Run("WithoutPerSecondRedis", testBasicConfig("8083", "false", "0", ""))
```
dweitzman, line 55: The first number in most of these `t.Run(...)` lines was the grpc port number. It's no longer needed since it's always 8083 now

test/integration/integration_test.go
```diff
@@ -45,239 +75,230 @@ func newDescriptorStatusLegacy(
...
+	s.LocalCacheSizeInBytes = localCacheSize
+	s.BackendType = "redis"
+	s.RedisUrl = "localhost:" + strconv.Itoa(redisPort)
+	s.RedisPerSecondUrl = "localhost:" + strconv.Itoa(perSecondPort)
+	return s
+}
+
 func TestBasicConfig(t *testing.T) {
 	common.WithMultiRedis(t, []common.RedisConfig{
-		{Port: 6379},
```
dweitzman, line 52: I also changed redis port 6379 to 6383 since 6379 is a traditional default port to run redis on and people might already have an instance running.

test/integration/integration_test.go
```diff
@@ -45,239 +75,230 @@ func newDescriptorStatusLegacy(
...
+	s.BackendType = "redis"
+
+	s.RuntimeWatchRoot = false
+
+	configRedisCluster(&s)
+
+	return testBasicBaseConfig(s)
 }
 
-func testBasicConfigWithoutWatchRootWithRedisSentinel(grpcPort, perSecond string, local_cache_size string) func(*testing.T) {
```
dweitzman, line 229: I'm noticing now that this helper function (`testBasicConfigWithoutWatchRootWithRedisSentinel`) was/is unused.

test/integration/integration_test.go
```diff
@@ -45,239 +75,230 @@ func newDescriptorStatusLegacy(
...
-	os.Setenv("REDIS_PERSECOND_TYPE", "single")
-
-	return testConfigReload(grpcPort, perSecond, local_cache_size)
+func testBasicConfigReload(perSecond bool, local_cache_size int, runtimeWatchRoot bool) func(*testing.T) {
+	s := makeSimpleRedisSettings(6383, 6380, perSecond, local_cache_size)
+	s.RuntimeWatchRoot = runtimeWatchRoot
+	return testConfigReload(s)
 }
 
-func testBasicConfigReloadWithRedisCluster(grpcPort, perSecond string, local_cache_size, runtimeWatchRoot string) func(*testing.T) {
```
dweitzman, line 255: This helper function was/is also unused

test/integration/integration_test.go
```diff
@@ -45,239 +75,230 @@ func newDescriptorStatusLegacy(
...
+	s.BackendType = "redis"
+
+	s.RuntimeWatchRoot = s.RuntimeWatchRoot
+
+	configRedisCluster(&s)
+
+	return testConfigReload(s)
 }
 
-func testBasicConfigReloadWithRedisSentinel(grpcPort, perSecond string, local_cache_size, runtimeWatchRoot string) func(*testing.T) {
```
dweitzman, line 271: And this one

src/server/server_impl.go
```diff
@@ -138,7 +140,20 @@ func (server *server) Start() {
...
+	server.httpListener = list
+	server.listenerMu.Unlock()
+	err = http.Serve(list, server.router)
+
+	server.listenerMu.Lock()
+	explicitlyStopped := server.stopped
+	server.listenerMu.Unlock()
+
+	if !explicitlyStopped {
+		logger.Fatal(err)
```
mattklein123, line 153: Just for my understanding, when does this happen? In the signal exit case, there is an exit(0) there, so that won't cause this, right? Is it basically just if somehow Serve() returns without us calling close()?

test/integration/integration_test.go
```diff
@@ -45,239 +75,230 @@ func newDescriptorStatusLegacy(
...
+		t.Run("WithoutPerSecondRedis", testBasicConfig(makeSimpleRedisSettings(6383, 6380, false, 0)))
+		t.Run("WithPerSecondRedis", testBasicConfig(makeSimpleRedisSettings(6383, 6380, true, 0)))
+		t.Run("WithoutPerSecondRedisWithLocalCache", testBasicConfig(makeSimpleRedisSettings(6383, 6380, false, 1000)))
+		t.Run("WithPerSecondRedisWithLocalCache", testBasicConfig(makeSimpleRedisSettings(6383, 6380, true, 1000)))
+		cacheSettings := makeSimpleRedisSettings(6383, 6380, false, 0)
+		cacheSettings.CacheKeyPrefix = "prefix:"
+		t.Run("WithoutPerSecondRedisWithCachePrefix", testBasicConfig(cacheSettings))
 	})
 }
-
```
mattklein123, line 62: nit: revert

mattklein123 (changes requested):
Thanks this is a great cleanup. One question and one small nit. 

Since you seem to be willing to do some tech debt cleanup, it would be great at some point if we could generate code coverage information that takes into account unit/integration tests. I haven't programmed any Go in years so not sure what the state of the art is there but if it's easy to setup it would be nice to see if we are missing any obvious coverage and if we break any coverage when we do major test refactors like this. Thank you!

/wait

src/server/server_impl.go
dweitzman: I looked at http/server.go and I can make this cleaner by checking if the err is http.ErrServerClosed. I'll do that

The other reasons for an error would be some kind of permanent network error from the listener

test/integration/integration_test.go
dweitzman: fixed

dweitzman:
> Thanks this is a great cleanup. One question and one small nit.
> 
> Since you seem to be willing to do some tech debt cleanup, it would be great at some point if we could generate code coverage information that takes into account unit/integration tests. I haven't programmed any Go in years so not sure what the state of the art is there but if it's easy to setup it would be nice to see if we are missing any obvious coverage and if we break any coverage when we do major test refactors like this. Thank you!
> 
> /wait

I'm curious how coverage work in go. If I have some time I'll poke around

dweitzman:
Force push was to fix DCO

mattklein123 (approved):
Thanks!

PR #220, opened 2021-02-03
Add EXTRA_TAGS config option to tag all stats

dweitzman:
This is useful for tagging all stats with things like build commit or host name.

Additionally:
- Reformat settings.go to make merge conflicts less common
- Factor out some common logic in integration_test.go
- Make integration_test.go slightly smarter & faster about
  waiting for test instances to come up instead of always waiting
  a full second.
- Upgrades the envconfig library to a version that supports map configs

### DIFF
--- a/src/settings/settings.go
+++ b/src/settings/settings.go
@@ -11,18 +11,34 @@ type Settings struct {
 	// runtime options
 	GrpcUnaryInterceptor grpc.ServerOption
 	// env config
-	Port                         int           `envconfig:"PORT" default:"8080"`
-	GrpcPort                     int           `envconfig:"GRPC_PORT" default:"8081"`
-	DebugPort                    int           `envconfig:"DEBUG_PORT" default:"6070"`
-	UseStatsd                    bool          `envconfig:"USE_STATSD" default:"true"`
-	StatsdHost                   string        `envconfig:"STATSD_HOST" default:"localhost"`
-	StatsdPort                   int           `envconfig:"STATSD_PORT" default:"8125"`
-	RuntimePath                  string        `envconfig:"RUNTIME_ROOT" default:"/srv/runtime_data/current"`
-	RuntimeSubdirectory          string        `envconfig:"RUNTIME_SUBDIRECTORY"`
-	RuntimeIgnoreDotFiles        bool          `envconfig:"RUNTIME_IGNOREDOTFILES" default:"false"`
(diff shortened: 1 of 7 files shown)


test/common/common.go
```diff
@@ -96,21 +96,8 @@ type MemcacheConfig struct {
 	Port int
 }
 
-// startCacheProcess starts memcache or redis as a subprocess and waits until the TCP port is open.
-func startCacheProcess(ctx context.Context, command string, args []string, port int) (context.CancelFunc, error) {
-	ctx, cancel := context.WithCancel(ctx)
-	cmd := exec.CommandContext(ctx, command, args...)
-
-	cmd.Stdout = os.Stdout
```
dweitzman, line 104: This is going to have a merge conflict with my other diff to fix the hangs on panic()

dweitzman: Merged

src/settings/settings.go
```diff
@@ -11,18 +11,34 @@ type Settings struct {
...
-	UseStatsd                    bool          `envconfig:"USE_STATSD" default:"true"`
-	StatsdHost                   string        `envconfig:"STATSD_HOST" default:"localhost"`
-	StatsdPort                   int           `envconfig:"STATSD_PORT" default:"8125"`
-	RuntimePath                  string        `envconfig:"RUNTIME_ROOT" default:"/srv/runtime_data/current"`
-	RuntimeSubdirectory          string        `envconfig:"RUNTIME_SUBDIRECTORY"`
-	RuntimeIgnoreDotFiles        bool          `envconfig:"RUNTIME_IGNOREDOTFILES" default:"false"`
-	RuntimeWatchRoot             bool          `envconfig:"RUNTIME_WATCH_ROOT" default:"true"`
-	LogLevel                     string        `envconfig:"LOG_LEVEL" default:"WARN"`
-	LogFormat                    string        `envconfig:"LOG_FORMAT" default:"text"`
+	Port      int `envconfig:"PORT" default:"8080"`
```
dweitzman, line 14: The reason for splitting settings.go into different groups is to reduce mere conflicts

Previously there were two reasons for merge conflicts:
- gofmt could reformat all the settings when a new setting is add
- people tend to add new settings to the end, which has a high likelihood of merge conflicts if multiple people are adding new settings

mattklein123 (approved):
Thanks!

PR #221, opened 2021-02-04
Fix tests hanging on panic() instead of crashing

dweitzman:
I don't claim to understand why, but sharing file descriptors with a subprocess
and then panic()ing seems to hang instead of exiting. The workaround is to
copy the output explicitly instead of sharing file descriptors.


mattklein123 (approved):
Thanks!

PR #225, opened 2021-02-07
docker_coverage build target

dweitzman:
Running "make docker_coverage" would create coverage.out and
suggest running "go tool cover --html=coverage.out" to explore it.

It would show the coverage of all unit and integration tests for
code under ./src/...

Other changes:
- "make docker_tests" no longer runs all the non-integration unit tests twice
- factors out the cache startup commands needed for integration tests into
  a script instead of the Makefile

### DIFF
--- a/Dockerfile.integration
+++ b/Dockerfile.integration
@@ -14,4 +14,5 @@ RUN make bootstrap
 
 COPY src /workdir/src
 COPY test /workdir/test
+COPY script /workdir/script
 CMD make tests_with_redis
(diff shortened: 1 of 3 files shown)


dweitzman:
Not sure how useful this is, but coverage came up as a topic in https://github.com/envoyproxy/ratelimit/pull/219#pullrequestreview-582017581 so I looked into it a bit and this might help someone who's actively curious about coverage

Not sure about what it would take to have coverage displayed on github in a useful way

mattklein123:
Thanks this is great. I think we have a couple of options:
1) Just merge as is and iterate
2) Actually run coverage and then upload the coverage HTML, etc. as an artifact in CI that can be explored.
3) Potentially gate CI on a coverage number.

WDYT?

/wait

PR #226, opened 2021-02-09
Add transport_api_version to http_filters config for ratelimit service

raflesw:
I found envoy was crashing with SEGV without this, on current version

### DIFF
--- a/examples/envoy/proxy.yaml
+++ b/examples/envoy/proxy.yaml
@@ -59,6 +59,7 @@ static_resources:
                         grpc_service:
                           envoy_grpc:
                             cluster_name: ratelimit
+                         transport_api_version: V3
                   - name: envoy.filters.http.router
                     typed_config: {}
                 route_config:


mattklein123:
Cc @htuch as this shouldn't crash envoy. This may have also been fixed on latest head? Either way this lgtm. Thanks!

mattklein123:
Oops sorry will need a dco fix. 

/wait

examples/envoy/proxy.yaml
```diff
@@ -59,6 +59,7 @@ static_resources:
                         grpc_service:
                           envoy_grpc:
                             cluster_name: ratelimit
+                         transport_api_version: V3
```
htuch, line 62: The indent here isn't right (one too many spaces) and it fails YAML parse. Do we have CI tests for this file?

htuch:
I tried this with Envoy @ HEAD on main and I don't see a SEGV without the extra `transport_api_version`.

mattklein123:
Done in https://github.com/envoyproxy/ratelimit/pull/227

PR #227, opened 2021-02-16
Add transport_api_version: V3 in example conf

fai555:
The docker-compose-example.yml crashes while sending a cURL request. After adding `transport_api_version: V3` in the `http_filters` spec for the `rate_limit_service` it started working.


mattklein123 (approved):
Thanks!

Issue #230, opened 2021-02-25
Proposal: Add WithinLimit Counter to RateLimitStats


jterryhao:
Hi team, I'm wondering if we can add a counter that counts the number of requests that is within limit (not over limit).

The reason is that we need to keep track how many requests are allowed by the rate limit service per second. We have some rate limits that don't allow any traffic and we want to monitor and make sure that no requests are going through. We are currently calculating it using
`rate(total_hits) - rate(over_limit)`

So for rate limits with 0 as the limit, this query should always return 0; however, prometheus' (we export the stastd metrics to prometheus) rate function does some smoothing/extrapolation and we often get non-zero numbers for those rate limits.

If we can add the WithinLimit counter, we can use that directly for monitoring allowed per second. Any help is appreciated. Thanks!

mattklein123:
Sure that sounds reasonable to me.

zufardhiyaulhaq:
I can take this one after the rolling window limit PR is merged if you do not mind.

jterryhao:
If you don't mind, I can also take this one if @zufardhiyaulhaq is busy with the rolling window limit PR (feels like a large one)

jterryhao:
Sorry I made a mistake in the example statsd exporter config, can I make a PR to fix this issue?

lmajercak-wish:
Can we close this?

PR #232, opened 2021-03-03
Add within limit counter

jterryhao:
Add WithinLimit Counter #230


lmajercak-wish:
can we also add it to [examples/prom-statsd-exporter/conf.yaml ](https://github.com/envoyproxy/ratelimit/blob/main/examples/prom-statsd-exporter/conf.yaml)?

mattklein123 (approved):
Thanks!

Issue #181, opened 2020-10-19
[Proposal] detailed metric mode, for value in descriptor KEY_VALUE pair, if value not explicitly defined/matched in config


wwillsey:
In order to gain greater observability of the operations of the rate limit service, a "detailed" metric mode that exposes metrics on the descriptor KEY_VALUE pairs in all cases, not only when the value is explicitly configured in the config.
These _values_ would allow for substantially more granular alerts / downstream reporting based on metrics emitted by the process when a descriptor key is matched, but its value is not.

While there's an understandable the concern over metric cardinality when including the values in the metrics, it is often the case that the possible values are well understood, and limited such that including them would pose little concern.

dweitzman:
Seems like a practical compromise might be to have a fix-sized buffer for each config that can hold a sample of real-world key/value pairs that were matched during particular time buckets, or what OpenCensus would call "exemplars" https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/Exemplars.md (also demonstrated in https://youtu.be/U72b4Nl0Ftw?t=1300)

mattklein123:
The exemplars approach sounds interesting. I would also be fine with an "all-metrics" mode as long as it is opt-in.

jespersoderlund:
I also have an interest in this and created an issue quite a while ago that turned "stale". I have now created an implementation of this, that would at least match our needs.

I added a key to the configuration called
"include_value_in_metric_when_not_specified"

which will override the default behavior and add the values to metrics.

I can create a PR attached to this issue and you can have a look

PR #233, opened 2021-03-10
Detailed Metrics

pradnic:
This PR adds _opt-in_ **detailed metrics** as discussed [here](https://github.com/envoyproxy/ratelimit/issues/181) This feature allows users to track totalhits, nearlimit, withinlimit, and overlimit on a per-descriptor basis, instead of on a per-configuration basis.

In order to activate this feature users must set DETAILED_METRICS_MODE environment variable to true at startup (default is false).

In order to support this feature we had to do some refactoring on the metric production.


pradnic:
Closing until we figure out DCO

pradnic:
Reopened https://github.com/envoyproxy/ratelimit/pull/233

Issue #234, opened 2021-03-11
Proposal: Enable emittance of total service call + response time metrics


thao-wish:
Currently ratelimit emits service error metrics that can be used to monitor error counts, but it doesn't emit total number of service calls or response time metrics that would enable us to monitor Requests Per Second/latency.

In our case, being able to monitor these metrics is an absolute necessity. Since the ratelimit service is used in production at lyft, I'm wondering if the team can share some insights on how RPS/latency metrics are measured at lyft. I can also work on a PR to enable these metrics.

dweitzman:
Stats does seems like an area that could use improvement. Not at lyft, but I can tell you what we've done.

For the moment we monitor latency and errors at the client side, with the rate limiter running as a sidecar on the same machine. The stats we use from the rate limiter itself are config loads and errors, memcache hit rates, per-descriptor stats, and a hacky total descriptor hit rate (which opentsdb is willing to do even though it's a aggregating across ton of metric names instead of something simple and easy like tags)

I briefly tested what happens when you get time/latency statistics from the `gostats` library and the output format wasn't compatible with our metrics agent. I think maybe it outputs floating point numbers and that wasn't accepted by the agent, or something like that (`statsd` is not a standardized protocol).

mattklein123:
> In our case, being able to monitor these metrics is an absolute necessity. Since the ratelimit service is used in production at lyft, I'm wondering if the team can share some insights on how RPS/latency metrics are measured at lyft. I can also work on a PR to enable these metrics.

We use Envoy :)

PR #236, opened 2021-03-12
Add server-side metrics

thao-wish:
Add server-side metrics #234

### DIFF
--- a/examples/prom-statsd-exporter/conf.yaml
+++ b/examples/prom-statsd-exporter/conf.yaml
@@ -71,7 +71,7 @@ mappings: # Requires statsd exporter >= v0.6.0 since it uses the "drop" action.
     name: "ratelimit_service_total_requests"
     match_metric_type: counter
   - match: "ratelimit_server.response_time"
-    name: "ratelimit_service_response_time_microseconds"
+    name: "ratelimit_service_response_time_seconds"
     timer_type: histogram
 
   - match: "ratelimit.service.config_load_success"
--- a/examples/prom-statsd-exporter/conf.yaml
+++ b/examples/prom-statsd-exporter/conf.yaml
@@ -67,12 +67,17 @@ mappings: # Requires statsd exporter >= v0.6.0 since it uses the "drop" action.
     labels:
       err_type: "$1"
 
-  - match: "ratelimit_server.total_requests"
+  - match: "ratelimit_server.*.total_requests"
     name: "ratelimit_service_total_requests"
     match_metric_type: counter
-  - match: "ratelimit_server.response_time"
+    labels:
+      grpc_method: "$1"
+
+  - match: "ratelimit_server.*.response_time"
     name: "ratelimit_service_response_time_seconds"
     timer_type: histogram
+    labels:
+      grpc_method: "$1"
 
   - match: "ratelimit.service.config_load_success"
     name: "ratelimit_service_config_load_success"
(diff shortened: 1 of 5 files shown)


examples/prom-statsd-exporter/conf.yaml
```diff
@@ -67,6 +67,13 @@ mappings: # Requires statsd exporter >= v0.6.0 since it uses the "drop" action.
     labels:
       err_type: "$1"
 
+  - match: "ratelimit_server.total_requests"
```
lmajercak-wish, line 70: I know we currently only have 1 API/handler, but it might make sense for these metrics to include the name of the API in it so that it can be easily extended in the future

examples/prom-statsd-exporter/conf.yaml
```diff
@@ -67,6 +67,13 @@ mappings: # Requires statsd exporter >= v0.6.0 since it uses the "drop" action.
...
+    name: "ratelimit_service_total_requests"
+    match_metric_type: counter
+  - match: "ratelimit_server.response_time"
+    name: "ratelimit_service_response_time_seconds"
```
lmajercak-wish, line 74: `s.responseTime.AddValue(float64(time.Since(start).Milliseconds()))` is this in milliseconds or seconds ?

lmajercak-wish, line 74: `s.responseTime.AddValue(float64(time.Since(start).Milliseconds()))` is this in milliseconds or seconds ?

thao-wish: According to the [statsd exporter](https://github.com/prometheus/statsd_exporter#statsd-timers-and-distributions) statsd format is milliseconds and prometheus format is seconds, so the exporter will convert the timer value to seconds.

examples/prom-statsd-exporter/conf.yaml
thao-wish: refactored to include name of gRPC method into the metrics

mattklein123 (changes requested):
Thanks looks good to me at a high level. Please add tests.

/wait

src/metrics/metrics.go
```diff
@@ -0,0 +1,45 @@
...
+
+// ServerReporter reports server-side metrics for ratelimit gRPC server
+type ServerReporter struct {
+	scope stats.Scope
+}
+
+func newServerMetrics(scope stats.Scope, fullMethod string) *serverMetrics {
+	_, methodName := splitMethodName(fullMethod)
+	ret := serverMetrics{}
+	ret.totalRequests = scope.NewCounter(methodName + ".total_requests")
```
dweitzman, line 23: I know the existing code tends to use metric names rather than tags, but for what its worth: I'd much rather see `methodName` as a tag rather than part of the metric name. (More generally, it seems like there's little to no reason a metric name ever needs to be dynamically created in a system that supports metrics with tags)

Using tags makes it simple in most metric systems to aggregate stats across method names or break down to specific subsets of method names.

Aggregating across metric names, by contrast, is not particularly simple or normal or standard in most metrics systems. Using different metric names also makes discovering what metric exist by looking at metric names a mess.

thao-wish: I agree that metrics tags are always the better choice than embedding in metrics name, but I don't think prom-statsd exporter supports the tagging format used by gostats. (please correct me if I'm wrong) So we'll have to do regexp matching to export it as a prom label, which is slower and more work than using glob patterns.

If there is a quick and easy way to export tags as prom labels, I'm happy to use tags instead.

mattklein123 (approved):
Thanks!

PR #237, opened 2021-03-12
Adds Detailed Metrics feature

pradnic:
This PR adds opt-in detailed metrics as discussed [here](https://github.com/envoyproxy/ratelimit/issues/181) This feature allows users to track totalhits, nearlimit, withinlimit, and overlimit on a per-descriptor basis, instead of on a per-configuration basis.

In order to activate this feature users must set DETAILED_METRICS_MODE environment variable to true at startup (default is false).

In order to support this feature we had to do some refactoring on the metric production.


mattklein123 (changes requested):
Thanks for working on this. At a high level this feature makes sense for opt-in, however this PR is too large. Can you please split this into a refactor PR and then a PR to add the new functionality? Thank you.

/wait

Issue #205, opened 2021-01-12
multiple memcached server support


bobo:
Im trying out the memcached support, and have trouble using more than one server.

having:
ENV MEMCACHE_HOST_PORT=host1t:11211,host2:11211,host3:11211
ENV BACKEND_TYPE=memcache

doesnt seem to work out of the box.
I looked into the memcache client used and it does support multiple servers (and handles key paritioning).

i tried the following change and then it seems to work atleast. 
But maybe im missing something and there is another way of configuring multiple memcache servers?

Index: src/memcached/cache_impl.go
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/memcached/cache_impl.go b/src/memcached/cache_impl.go
--- a/src/memcached/cache_impl.go	(revision 6f5de117a9e59e80f3dbb05ad8c47d1cea85b87f)
+++ b/src/memcached/cache_impl.go	(date 1610458313644)
@@ -19,6 +19,7 @@
 	"context"
 	"math/rand"
 	"strconv"
+	"strings"
 	"sync"
 
 	"github.com/coocood/freecache"
@@ -186,7 +187,7 @@
 func NewRateLimitCacheImplFromSettings(s settings.Settings, timeSource utils.TimeSource, jitterRand *rand.Rand,
 	localCache *freecache.Cache, scope stats.Scope) limiter.RateLimitCache {
 	return NewRateLimitCacheImpl(
-		memcache.New(s.MemcacheHostPort),
+		memcache.New(strings.Split(s.MemcacheHostPort,",")...),
 		timeSource,
 		jitterRand,
 		s.ExpirationJitterMaxSeconds,

dweitzman:
That seems reasonable to add

Memcache support is quite recent. I didn't add multiple host support in the initial version since we're running mcrouter on localhost to handle cluster management

PR #238, opened 2021-03-21
Support multiple memcache hosts in memcache backend

nezdolik:
Fixes [#205](https://github.com/envoyproxy/ratelimit/issues/205)

Signed-off-by: Kateryna Nezdolii <nezdolik@spotify.com>


nezdolik:
cc @mattklein123 @dweitzman

src/settings/settings.go
```diff
@@ -58,7 +58,7 @@ type Settings struct {
 	RedisPerSecondPipelineLimit  int           `envconfig:"REDIS_PERSECOND_PIPELINE_LIMIT" default:"0"`
 
 	// Memcache settings
-	MemcacheHostPort string `envconfig:"MEMCACHE_HOST_PORT" default:""`
+	MemcacheUrl string `envconfig:"MEMCACHE_URL" default:""`
```
dweitzman, line 61: You should be able to change the type here to `[]string` and avoid the need to call `strings.Split` later

https://github.com/kelseyhightower/envconfig has some detail about what the envconfig library supports

README.md
```diff
@@ -577,7 +577,7 @@ Experimental Memcache support has been added as an alternative to Redis in v1.5.
 
 To configure a Memcache instance use the following environment variables instead of the Redis variables:
 
-1. `MEMCACHE_HOST_PORT=<host:port>`
+1. `MEMCACHE_URL=`: a comma separated list of hostname:port pairs for memcache nodes.
```
dweitzman, line 580: I wasn't sure exactly how this behaves in the client library. Seems like the default behavior if you use a list of servers and not a custom selector is at https://github.com/bradfitz/gomemcache/blob/a41fca850d0b/memcache/selector.go#L114: it calculates ChecksumIEEE for a particular key has a hash into the list of backends.

Might be worth noting somewhere here in the documentation that when using a list of memcache notes the same list should be used on all rate limiters because backend selection is done based on the list and the key checksum.

nezdolik: Right, same key should be hashed to same memcache node. That's a great point, will mention in docs.

dweitzman:
Looks good to me. The only minor thing which isn't super important to me is that the change from MEMCACHE_HOST_PORT to MEMCACHE_URL is backwards incompatible and doesn't seem strictly necessary. I'm not sure that a `host:port` would really be considered a URL since there's no scheme?...

I don't really have strong feelings, though.

mattklein123:
> I don't really have strong feelings, though.

If there is a way to not break backwards compat that would be much preferred? Can we do that? Thank you.

/wait

PR #246, opened 2021-04-26
Add env var to configure max idle connections per memcache node

petedmarsh:
The underlying memcache client library allows this to be configured,
and currently defaults to a value of 2, see:

https://github.com/bradfitz/gomemcache/blob/master/memcache/memcache.go#L72
https://github.com/bradfitz/gomemcache/blob/master/memcache/memcache.go#L145
https://github.com/bradfitz/gomemcache/blob/master/memcache/memcache.go#L239

This allows this value to be configured by a new environmet variable:

MEMCACHE_MAX_IDLE_CONNS

which defaults to -1 meaning the default from the library will apply
(which is the current behaviour).

Signed-off-by: Peter Marsh <pete.d.marsh@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -585,6 +585,7 @@ To configure a Memcache instance use the following environment variables instead
 1. `MEMCACHE_HOST_PORT=`: a comma separated list of hostname:port pairs for memcache nodes.
 1. `BACKEND_TYPE=memcache`
 1. `CACHE_KEY_PREFIX`: a string to prepend to all cache keys
+1. `MEMCACHE_MAX_IDLE_CONNS=2`: the maximum number of idle TCP connections per memcache node, `2` is the default of the underlying library
 
 With memcache mode increments will happen asynchronously, so it's technically possible for
 a client to exceed quota briefly if multiple requests happen at exactly the same time.
--- a/src/settings/settings.go
+++ b/src/settings/settings.go
@@ -68,6 +68,12 @@ type Settings struct {
 
 	// Memcache settings
 	MemcacheHostPort []string `envconfig:"MEMCACHE_HOST_PORT" default:""`
+	// MemcacheMaxIdleConns sets the maximum number of idle TCP connections per memcached node.
+	// The default is 2 as that is the default of the underlying library. This is the maximum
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -585,6 +585,7 @@ To configure a Memcache instance use the following environment variables instead
 1. `MEMCACHE_HOST_PORT=`: a comma separated list of hostname:port pairs for memcache nodes.
 1. `BACKEND_TYPE=memcache`
 1. `CACHE_KEY_PREFIX`: a string to prepend to all cache keys
+1. `MEMCACHE_MAX_IDLE_CONNS=-1`: the maximum number of idle TCP connections per memcache node, if `-1` this will use the default of the underlying library (currently 2)
```
nezdolik, line 588: This could be confusing for users that setting value to `-1` translates to `2`. Can we instead allows users to configure values within [1, N] range and internally check if this variable is not set, then don't propagate it to memcache client

src/settings/settings.go
```diff
@@ -68,6 +68,9 @@ type Settings struct {
 
 	// Memcache settings
 	MemcacheHostPort []string `envconfig:"MEMCACHE_HOST_PORT" default:""`
+	// MemcacheMaxIdleConns sets the maximum numebr of idle TCP connections per memcached node.
```
nezdolik, line 71: nit: /s/numerb/number

README.md
nezdolik: Could you explain why users may need to tune this parameter?

nezdolik (commented):
Please add integration tests

petedmarsh:
@nezdolik I think I've addressed your comments, except one: I couldn't see any other the other env vars being validated before use, I am happy to put that in if you still think I should but I'm not sure where is best to do so?

nezdolik:
@petedmarsh there is indeed no consistent way across codebase to validate env variables. This could be a subject of separate improvement task. I've checked code of memcache client, in case `max_idle_connections` is passed as any value below 2, client would fall back to reasonable default.

nezdolik:
@mattklein123 this is ready for review

mattklein123 (approved):
Thanks!

Issue #250, opened 2021-05-03
High CPU usage with GRPC + Memcache on 32 cores machine


storozhukBM:
I noticed that GRPC + Memcache configuration usage leads to unexpectedly high CPU usage on 32 cores machine even with moderately low 5000 req/sec. Curiously this is not the case on machines with smaller CPU count or if you manually set lower GOMAXPROCS. Also, lower GOMAXPROCS of 16 and 8 doesn't lead to lower latency on this 5k req/sec rate, just overall CPU usage on the machine goes down.

storozhukBM:
It's fixed and already merged to main branch

PR #251, opened 2021-05-03
Reduce short living tasks produced by memcached implementation and move them to goroutines pool

storozhukBM:
During investigation on https://github.com/envoyproxy/ratelimit/issues/250 I made a quick look at pprof results didn't show any noticeable hot spots in the `github.com/envoyproxy/ratelimit` or `github.com/bradfitz/gomemcache`
but a lot of samples were in runtime.schedule, runtime.findrunnable, runtime.futex. So I decided to reduce the potential load on the scheduler by moving the continuous spawn of new short-lived goroutines to a pool that can automatically expand to a necessary size, stabilize and stop creating new goroutines if the load remains stable.

This simple change reduced overall CPU consumption on 32 cores machine by 20% in the same setup that is described in the issue https://github.com/envoyproxy/ratelimit/issues/250

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.16 AS build
+FROM golang:1.14 AS build
 WORKDIR /ratelimit
 
 ENV GOPROXY=https://proxy.golang.org
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.14 AS build
+FROM golang:1.16 AS build
 WORKDIR /ratelimit
 
 ENV GOPROXY=https://proxy.golang.org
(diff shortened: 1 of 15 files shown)


mattklein123 (dismissed):
Makes sense, thanks!

mattklein123:
Oops looks like a real CI issue, please take a look.

/wait

mattklein123:
I would be fine with bumping the Go version, please go for it!

/wait

storozhukBM:
@mattklein123 I bumped Go version to 1.16, please take a look

mattklein123:
I think there might be another fix required due to the Go update. Sorry. :(

/wait

mattklein123 (approved):
Thanks!

PR #253, opened 2021-05-06
Add support for SRV records for memcache clients

petedmarsh:
This allows MEMCAHE_SRV to be specified as an SRV record from
which multiple memcache hosts can be resolved. For example:

    MEMCACHE_SRV=_memcache._tcp.mylovelydomain.com

This can be used instead of MEMCACHE_HOST_PORT.

This will then be resolved and whatever set of servers it represents
will be used as the set of memcache servers to connect to. At this
stage neither priority or weight is supported, though weight could
be fairly straightforwardly in future.

The SRV can be polled periodically for new servers by setting the
following env var (with 0 meaning "never check"):

    MEMCACHE_SRV_REFRESH=600s # supports standard go durations

### DIFF
--- a/src/settings/settings.go
+++ b/src/settings/settings.go
@@ -76,7 +76,9 @@ type Settings struct {
 	// number of connections to memcache kept idle in pool, if a connection is needed but none
 	// are idle a new connection is opened, used and closed and can be left in a time-wait state
 	// which can result in high CPU usage.
-	MemcacheMaxIdleConns int `envconfig:"MEMCACHE_MAX_IDLE_CONNS" default:"2"`
+	MemcacheMaxIdleConns int           `envconfig:"MEMCACHE_MAX_IDLE_CONNS" default:"2"`
+	MemcacheSrv          string        `envconfig:"MEMCACHE_SRV" default:""`
+	MemcacheSrvRefresh   time.Duration `envconfig:"MEMCACHE_SRV_REFRESH" default:"0"`
 }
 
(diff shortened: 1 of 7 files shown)


petedmarsh:
I have run this myself and confirmed it works, but there are elements I have struggled to lock down with unit tests, namely:

1. An integration test using an SRV
2. Tests for periodic refreshing (I'm not sure how to manage the clock)

My knowledge of go is quite limited so I can happily add such tests if they are deemed necessary and I can get some pointers on how to do it :)

mattklein123:
Please merge main.

/wait

src/settings/settings.go
```diff
@@ -62,7 +62,7 @@ type Settings struct {
 	RedisPerSecondPoolSize   int    `envconfig:"REDIS_PERSECOND_POOL_SIZE" default:"10"`
 	RedisPerSecondAuth       string `envconfig:"REDIS_PERSECOND_AUTH" default:""`
 	RedisPerSecondTls        bool   `envconfig:"REDIS_PERSECOND_TLS" default:"false"`
-	// RedisPerSecondPipelineWindow sets the duration after which internal pipelines will be flushed for per second redis.
+	// RedisPerSec	ondPipelineWindow sets the duration after which internal pipelines will be flushed for per second redis.
```
petedmarsh, line 65: I'll fix this

mattklein123 (approved):
Thanks!

Issue #256, opened 2021-05-25
Flaky test: ratelimit_legacy_test.go:157: TestServiceLegacy


sunjayBhatia:
Failing in CI but passes when I run locally

See runs:
- https://github.com/envoyproxy/ratelimit/runs/2667881508
- https://github.com/envoyproxy/ratelimit/runs/2667889432?check_suite_focus=true

```
--- FAIL: TestServiceLegacy (0.00s)
    ratelimit_legacy_test.go:157: 
        	Error Trace:	ratelimit_legacy_test.go:157
        	Error:      	Not equal: 
        	            	expected: int(1)
        	            	actual  : uint64(0x0)
        	Test:       	TestServiceLegacy
time="2021-05-25T17:35:39Z" level=error msg="error loading new configuration from runtime: load error"
time="2021-05-25T17:35:39Z" level=error msg="error loading new configuration from runtime: load error"
time="2021-05-25T17:35:39Z" level=error msg="error loading new configuration from runtime: load error"
time="2021-05-25T17:35:39Z" level=error msg="error loading new configuration from runtime: load error"
FAIL
FAIL	github.com/envoyproxy/ratelimit/test/service	0.067s
FAIL
```

Test: https://github.com/envoyproxy/ratelimit/blob/bc0e9faec3397a6dad9b27fc983be8a229306ff6/test/service/ratelimit_legacy_test.go#L157

PR #258, opened 2021-05-25
Fix flaky test TestServiceLegacy

sunjayBhatia:
User deferred barrier.signal() so panic definitely occurs before
we continue on in test.

Config reload uses recover() and increments config load counter, tests
were failing to see config load error counter increment.

Fixes: #256


mattklein123 (approved):
Thanks!

Issue #241, opened 2021-03-31
Proposal: Add explicit unlimited rate limits


thao-wish:
Currently we can achieve implicit unlimited rate limits by not specifying a rate limit in the configuration. However, we'll lose the metrics (total_hits, within_limits) for those rate limits. The proposal is to add an explicit unlimited rate limit. An example config file could look like this:
```
domain: mongo_cps
descriptors:
  - key: database
    value: users
    rate_limit:
      unit: second
      requests_per_unit: unlimited
```
We can skip the trip to redis/memcached if unlimited rate limits are requested. Please let me know if the team is open to this request or discuss any alternatives. The main concern is to be able to get the metrics for the unlimited rate limits.

mattklein123:
IMO I would rather just implement https://github.com/envoyproxy/ratelimit/issues/218 which I think covers this case?

thao-wish:
> IMO I would rather just implement #218 which I think covers this case?

I think it's a bit different. Please correct me if I'm wrong.

Test limits are processed in the same way as real limits, so they still have to go through redis/memcached. Currently the unlimited rate limits do not go to redis/memcached, so we want the new unlimited rate limits to be the same. It's also good for performance.

mattklein123:
Yeah fair enough. I agree we can look at this independently.

lmajercak-wish:
Hey @mattklein123, I would like to pick this up. 

I've already looked into this briefly and it seems like the biggest issue here will be deciding what to return for rate limits which are set to 'unlimited'. Specifically, we have the `uint32 requests_per_unit` and `uint32 limit_remaining` fields.

Any suggestions on what we do with these? If we omitted them, then I think there isn't a difference between a response for an unlimited ratelimit and one that does not exist, which might cause confusion. Alternatively, I guess we could return something like uint32 max value, but that does not seem ideal either.

lmajercak-wish:
I just realized that, if the limit is set to 0, these two fields are omitted. I guess we could do the same for unlimited ones then?

lmajercak-wish:
Opened https://github.com/envoyproxy/ratelimit/pull/261

zakhenry:
This appears to be implemented now, can this issue be closed?

PR #261, opened 2021-06-09
Explicit unlimited ratelimits

lmajercak-wish:
Provide the ability to explicitly set a rate limit descriptor to be unlimited to avoid calls to redis/memcached.

https://github.com/envoyproxy/ratelimit/issues/241


examples/ratelimit/config/example.yaml
```diff
@@ -27,3 +27,13 @@ descriptors:
         rate_limit:
           unit: second
           requests_per_unit: 1
+      - key: bay
+        rate_limit:
+          unit: second
+          requests_per_unit: 5
```
mattklein123, line 33: Does specifying unit/requests_per_unit make sense? Probably not? Also, can you document the feature in README.md?

mattklein123 (changes requested):
Thanks for adding this. Can you add some documentation and then we can go from there? Thank you.

/wait

examples/ratelimit/config/example.yaml
lmajercak-wish: Thanks for looking at this @mattklein123. Removed `requests_per_unit`. The `unit` cannot be removed right now, it complains about it being of an invalid format. Do we wanna make it so that we can omit the `unit` too?

Also added an example to the README.

README.md
```diff
@@ -329,6 +330,26 @@ descriptors:
...
+#### Example 5
+
+We can also define unlimited rate limit descriptors:
+
+```yaml
+domain: internal
+descriptors:
+  - key: ldap
+    rate_limit:
+      unit: second
```
mattklein123, line 342: Is there a way we can remove the requirement to specify `unit`? It doesn't make sense in the unlimited context. Or perhaps: `unit: unlimited` if that is easier?

mattklein123 (commented):
/wait

README.md
lmajercak-wish: Changed it so that specifying unit for unlimited descriptors is not needed

mattklein123:
Sorry can you fix DCO? Feel free to force push in this case and then we can continue the review from there.

/wait

lmajercak-wish:
> Sorry can you fix DCO? Feel free to force push in this case and then we can continue the review from there.
> 
> /wait

done

lmajercak-wish:
Hi @mattklein123, would you be able to take a look again? Thanks! :)

src/config/config_impl.go
```diff
@@ -112,19 +115,21 @@ func (this *rateLimitDescriptor) loadDescriptors(config RateLimitConfigToLoad, p
 		var rateLimit *RateLimit = nil
 		var rateLimitDebugString string = ""
 		if descriptorConfig.RateLimit != nil {
+			unlimited := descriptorConfig.RateLimit.Unlimited
+
 			value, present :=
 				pb.RateLimitResponse_RateLimit_Unit_value[strings.ToUpper(descriptorConfig.RateLimit.Unit)]
-			if !present || value == int32(pb.RateLimitResponse_RateLimit_UNKNOWN) {
+			if (!present || value == int32(pb.RateLimitResponse_RateLimit_UNKNOWN)) && !unlimited {
```
mattklein123, line 122: Can you split this logic out so that we actually check if unlimited is set with a unit at all and have an error message (and test) for that? I think that would be more clear to the user.

src/memcached/cache_impl.go
```diff
@@ -83,6 +83,12 @@ func (this *rateLimitMemcacheImpl) DoLimit(
 			continue
 		}
 
+		// Check if the key is unlimited
```
mattklein123, line 86: It seems odd that we have to hit the cache at all for unlimited limits? Can this be avoided somehow?

mattklein123 (changes requested):
Thanks LGTM with small comments.

/wait

src/memcached/cache_impl.go
lmajercak-wish: Moved the logic to ratelimit.go

test/service/ratelimit_test.go
```diff
@@ -240,3 +240,40 @@ func TestInitialLoadError(test *testing.T) {
...
+
+	response, err := service.ShouldRateLimit(nil, request)
+	common.AssertProtoEqual(
+		t.assert,
+		&pb.RateLimitResponse{
+			OverallCode: pb.RateLimitResponse_OK,
+			Statuses: []*pb.RateLimitResponse_DescriptorStatus{
+				{Code: pb.RateLimitResponse_OK, CurrentLimit: limits[0].Limit, LimitRemaining: 9},
+				{Code: pb.RateLimitResponse_OK, CurrentLimit: nil, LimitRemaining: 0},
+				{Code: pb.RateLimitResponse_OK, CurrentLimit: nil, LimitRemaining: 0},
```
lmajercak-wish, line 275: @mattklein123 this is something I'm concerned with. We cannot distinguish between a descriptor that wasn't found and a descriptor that's unlimited :(. This will make it hard for the clients to decide what to do. Any ideas how we can fix this without changing the proto spec?

README.md
```diff
@@ -329,6 +330,25 @@ descriptors:
...
+    rate_limit:
+      unlimited: true
+
+  - key: azure
+    rate_limit:
+      unit: minute
+      requests_per_unit: 100
+```
+
+For an unlimited descriptor, the request will not be sent to the underlying cache (Redis/Memcached), but will be quickly returned locally by the ratelimit instance.
```
mattklein123, line 350: Can you mention why someone would want to do this? I assume for stats, etc.?

src/config/config_impl.go
mattklein123: (Ping on this)

src/memcached/cache_impl.go
```diff
@@ -136,7 +136,7 @@ func (this *rateLimitMemcacheImpl) increaseAsync(cacheKeys []limiter.CacheKey, i
 	limits []*config.RateLimit, hitsAddend uint64) {
 	defer this.waitGroup.Done()
 	for i, cacheKey := range cacheKeys {
-		if cacheKey.Key == "" || isOverLimitWithLocalCache[i] {
+		if cacheKey.Key == "" || isOverLimitWithLocalCache[i] || limits[i].Unlimited {
```
mattklein123, line 139: Is this still needed if we lift this out of the cache layer?

test/service/ratelimit_test.go
mattklein123: Yeah good point. It's not perfect but in the unlimited case why don't we set LimitRemaining to max uint? We can document that. In a perfect world we would add a boolean to the proto but I'm not sure it's worth it right now?

mattklein123 (changes requested):
/wait

README.md
lmajercak-wish: Well, stats yes. But also, in our case, our client will not allow a request to go through if a descriptor is not found, so we do have to have it defined.

src/memcached/cache_impl.go
lmajercak-wish: Removing

README.md
lmajercak-wish: I'll add something to the readme

lmajercak-wish: Added

test/service/ratelimit_test.go
lmajercak-wish: Yea that sounds good to me, changed to max.uint32_t

src/config/config_impl.go
lmajercak-wish: I wasn't too sure what exactly you meant, so please review this again :)

mattklein123:
Please do not force push after the PR is started. It breaks the review process. Just merge main and add commits. Thank you!

src/service/ratelimit.go
```diff
@@ -118,9 +135,16 @@ func (this *service) shouldRateLimitWorker(
 	response.Statuses = make([]*pb.RateLimitResponse_DescriptorStatus, len(request.Descriptors))
 	finalCode := pb.RateLimitResponse_OK
 	for i, descriptorStatus := range responseDescriptorStatuses {
-		response.Statuses[i] = descriptorStatus
-		if descriptorStatus.Code == pb.RateLimitResponse_OVER_LIMIT {
-			finalCode = descriptorStatus.Code
+		if isUnlimited[i] {
+			response.Statuses[i] = &pb.RateLimitResponse_DescriptorStatus{
+				Code:           pb.RateLimitResponse_OK,
+				LimitRemaining: math.MaxUint32,
```
mattklein123, line 141: Can you document this somewhere? It should probably go directly in the proto docs as a comment? Or potentially also somewhere in the README?

mattklein123 (changes requested):
Thanks LGTM with a small doc comment.

/wait

lmajercak-wish:
> Please do not force push after the PR is started. It breaks the review process. Just merge main and add commits. Thank you!

I just followed the DCO steps

src/service/ratelimit.go
lmajercak-wish: Not sure what you mean by the proto docs, I added a comment to the README

mattklein123: I mean doing a PR here also: https://github.com/envoyproxy/envoy/blob/7c1991ffe382bacf2bf415a643c5ea7428d0880e/api/envoy/service/ratelimit/v3/rls.proto#L130-L131

This is the canonical service definition for the ratelimit service.

mattklein123:
@lmajercak-wish now that DCO is broken you will need to squash and force push, so please do that. In the future please make sure to only add commits and sign off each commit. Thank you!

/wait

mattklein123 (dismissed):
Thanks!

mattklein123:
Oops looks like CI issue.

/wait

mattklein123 (approved):
Thank you!

lmajercak-wish:
> Thank you!

Thank you @mattklein123 !

PR #266, opened 2021-06-30
Add memcached error stats

devincd:

### DIFF
--- a/src/service/ratelimit.go
+++ b/src/service/ratelimit.go
@@ -2,14 +2,16 @@ package ratelimit
 
 import (
 	"fmt"
-	"github.com/envoyproxy/ratelimit/src/stats"
 	"strings"
 	"sync"
+
+	"github.com/envoyproxy/ratelimit/src/stats"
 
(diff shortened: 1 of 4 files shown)


src/service/ratelimit.go
```diff
@@ -151,6 +153,12 @@ func (this *service) ShouldRateLimit(
 				this.stats.ShouldRateLimit.ServiceError.Inc()
 				finalError = t
 			}
+		case memcached.MemcacheError:
+			{
+				this.stats.ShouldRateLimit.MemcachedError.Inc()
+				finalError = t
+			}
+
```
mattklein123, line 161: I'm not crazy about this fact that we are leaking the cache implementation abstraction to this layer. Can you potentially refactor this so redis/memcache/etc. errors are handled internal to the implementation or there is some other way of hiding which counter needs incrementing?

mattklein123 (changes requested):
Thanks for improving the stats. Small comment.

/wait

PR #273, opened 2021-07-28
Fix bug and code optimization

devincd:

If redis url is unix domain socket, will panic. The log is bellow:
```
WARN[0000] connecting to redis on /var/run/nutcracker/ratelimit.sock with pool size 10
panic: dial tcp: address /var/run/nutcracker/ratelimit.sock: missing port in address
```

### DIFF
--- a/src/server/server_impl.go
+++ b/src/server/server_impl.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"expvar"
 	"fmt"
-	"github.com/envoyproxy/ratelimit/src/stats"
 	"io"
 	"net/http"
 	"net/http/pprof"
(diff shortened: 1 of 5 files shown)


src/server/server_impl.go
```diff
@@ -300,15 +301,6 @@ func (server *server) handleGracefulShutdown() {
...
-		server.grpcServer.GracefulStop()
-		server.listenerMu.Lock()
-		if server.debugListener.listener != nil {
-			server.debugListener.listener.Close()
-		}
-		if server.httpServer != nil {
-			server.httpServer.Close()
-		}
-		server.listenerMu.Unlock()
-		os.Exit(0)
```
mattklein123, line 312: I think we should leave the `Exit(0)` call after the `Stop()` call? Is it not needed?

mattklein123 (commented):
Nice, thanks. One question.

/wait-any

src/server/server_impl.go
devincd: yes. it is needed

mattklein123 (approved):
Thanks!

PR #281, opened 2021-08-06
Idea: Support of regex functionality

mgnisia:
To explain the workflow of the incoming request I wrote the following pseudo code: 

image

The main idea is that the incoming descriptor has only a single entry and if the entry is matching the regex expression, the request will be modified such that the value of the redis keys can be checked. I know that there has been a similar issue but the we couldn't find an appropriate solution based on with a specific descriptor.

### DIFF
--- a/src/config/config_impl.go
+++ b/src/config/config_impl.go
@@ -241,6 +241,7 @@ func (this *rateLimitConfigImpl) Dump() string {
 func (this *rateLimitConfigImpl) GetLimit(
 	ctx context.Context, domain string, descriptor *pb_struct.RateLimitDescriptor) *RateLimit {
 
+	logger.Debugf("Descriptor %v", descriptor)
 	logger.Debugf("starting get limit lookup")
 	var rateLimit *RateLimit = nil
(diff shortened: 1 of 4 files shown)


ysawa0:
I took a quick look. While I think a feature like this would be useful, if there's additional regex checks being done on every request there could be performance implications; so it may need to be gated behind some env var flag.

src/config/config_impl.go
```diff
@@ -241,6 +241,7 @@ func (this *rateLimitConfigImpl) Dump() string {
 func (this *rateLimitConfigImpl) GetLimit(
 	ctx context.Context, domain string, descriptor *pb_struct.RateLimitDescriptor) *RateLimit {
 
+	logger.Debugf("Descriptor %v", descriptor)
```
mattklein123, line 244: nit: merge this into the following log line

mattklein123 (changes requested):
Thanks for the contribution. I agree with @ysawa0 that this should be configuration gated somehow. It also needs documentation. Thank you!

/wait

PR #282, opened 2021-08-06
Add a little code optimization

devincd:


mattklein123 (changes requested):
Thanks, are there any tests for the client? Can we possibly add some? Also are there docs about the client in README? Can we add some for the new parameter?

/wait

PR #288, opened 2021-09-07
Add configurable GrpcMaxConnectionAge, GrpcMaxConnectionAgeGrace

lmajercak-wish:
Server-side configuration of MaxConnectionAge, MaxConnectionAgeGrace is needed to achieve better load balancing behavior in auto-scaled scenarios: https://github.com/grpc/grpc/issues/12295#issuecomment-382794204

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@
       - [Example 5](#example-5)
   - [Loading Configuration](#loading-configuration)
   - [Log Format](#log-format)
+  - [GRPC Keepalive](#grpc-keepalive)
 - [Request Fields](#request-fields)
 - [GRPC Client](#grpc-client)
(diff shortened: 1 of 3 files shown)


ysawa0:
Changes makes sense to me. Could you add docs about the new settings in the README and maybe some simple tests? Thanks! 
/wait

lmajercak-wish:
> Changes makes sense to me. Could you add docs about the new settings in the README and maybe some simple tests? Thanks!
> /wait

Added to the README. What are you thinking for the unit test?

ysawa0 (approved):

ysawa0:
I was going to suggest adding a simple integration test but it's already being tested via the default values so I guess this is fine. The defaults look sane as well. Will go ahead and merge, thanks!

PR #289, opened 2021-09-07
Configurable Redis TLS Config through settings

JavierR14:
At the moment, the redis tls config that it's using is a default and there is no way to change it. This makes it so that clients can specify their own TLS config if they choose to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -542,8 +542,6 @@ As well Ratelimit supports TLS connections and authentication. These can be conf
 1. `REDIS_AUTH` & `REDIS_PERSECOND_AUTH`: set to `"password"` to enable authentication to the redis host.
 1. `CACHE_KEY_PREFIX`: a string to prepend to all cache keys
 
-Ratelimit also provides the ability to specify TLS settings via `RedisTlsConfig` in the [settings](https://github.com/envoyproxy/ratelimit/blob/master/src/settings/settings.go).
-
 ## Redis type
 
 Ratelimit supports different types of redis deployments:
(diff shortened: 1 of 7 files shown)


JavierR14:
@envoyproxy/ratelimit-maintainers

mattklein123:
Thanks, please add documentation in the README and also check CI.

/wait

JavierR14:
I can't seem to get integration tests working locally even on the main branch for envoyproxy. Not sure how to kick off CI myself on here.

README.md
```diff
@@ -542,6 +542,8 @@ As well Ratelimit supports TLS connections and authentication. These can be conf
 1. `REDIS_AUTH` & `REDIS_PERSECOND_AUTH`: set to `"password"` to enable authentication to the redis host.
 1. `CACHE_KEY_PREFIX`: a string to prepend to all cache keys
 
+Ratelimit also provides the ability to specify TLS settings via `RedisTlsConfig` in the [settings](https://github.com/envoyproxy/ratelimit/blob/master/src/settings/settings.go).
```
mattklein123, line 545: Sorry it's not clear to me how this is used. Does someone have to instantiate this by code? Or can this be set via env variables?

/wait

JavierR14: Someone has to instantiate it by code. You shouldn't be setting a TLS config through env variables as it would require dealing with certs and key files and needs code. TLS configs should be configurable and this allows the client using ratelimit to configure it if needed.

mattklein123: OK, then I guess this isn't a general purpose feature so I'm not sure it's worth documenting like this. It would probably be better long term to allow the config to be read from a YAML file of some type. I would remove this from the README and put a TODO in the settings code to have that setting be out of box user configurable somehow.

/wait

JavierR14: Done. I've added a TODO line. I do think providing an out of the box user configurable setting is a good idea but ultimately will probably not be used much. Setting up a Redis TLS config will be very different from case to case and most of the time users will provide it through code.

mattklein123 (dismissed):
Thanks!

mattklein123:
Looks like it needs a format fix.

/wait

mattklein123 (approved):

Issue #291, opened 2021-09-19
Support adding custom headers


jespersoderlund:
Support for adding custom header names for the ratelimit headers.

The Envoy implementation is not compatible with the ratlimit draft spec for example, it's using "X-Ratelimit-" prefix though the draft standard uses just "Ratelimit-" prefix.

It's also useful to be able to specify a specific prefix if it's an API that you where you want to have the control of the specification yourself

jespersoderlund:
thanks

PR #292, opened 2021-09-19
Addressing issue #291 - Adding custom headers with the ratelimit trig…

jespersoderlund:
Issue #291 

Adding custom headers with rate limit data so that there is the option to use the draft specification specified headers of "Ratelimit-" instead of "X-Ratelimit-" or other custom naming schemes to conform to corporate standards

### DIFF
--- a/README.md
+++ b/README.md
@@ -625,6 +625,14 @@ descriptors will fail. Descriptors sent to Memcache should not contain whitespac
 When using multiple memcache nodes in `MEMCACHE_HOST_PORT=`, one should provide the identical list of memcache nodes
 to all ratelimiter instances to ensure that a particular cache key is always hashed to the same memcache node.
 
+# Custom headers
+Ratelimit service can be configured to return custom headers with the ratelimit information.
+
+Setting _all_ the following environment variables to the header name to use:
+1. `LIMIT_LIMIT_HEADER` - The value will be the current limit value closest to being triggered, currently the optional quota policies are not added
+1. `LIMIT_REMAINING_HEADER` - The value will be the remaining quota
+1. `LIMIT_RESET_HEADER` - The value will be the number of seconds until the limit is being reset
+
 # Contact
 
 * [envoy-announce](https://groups.google.com/forum/#!forum/envoy-announce): Low frequency mailing
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@
   - [One Redis Instance](#one-redis-instance)
   - [Two Redis Instances](#two-redis-instances)
 - [Memcache](#memcache)
+- [Custom headers](#custom-headers)
 - [Contact](#contact)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
--- a/README.md
+++ b/README.md
@@ -627,12 +627,13 @@ When using multiple memcache nodes in `MEMCACHE_HOST_PORT=`, one should provide 
 to all ratelimiter instances to ensure that a particular cache key is always hashed to the same memcache node.
 
 # Custom headers
-Ratelimit service can be configured to return custom headers with the ratelimit information. It will populate the response_headers_to_add as part of the [RateLimitRespons](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/ratelimit/v3/rls.proto#service-ratelimit-v3-ratelimitresponse).
+Ratelimit service can be configured to return custom headers with the ratelimit information. It will populate the response_headers_to_add as part of the [RateLimitResponse](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/ratelimit/v3/rls.proto#service-ratelimit-v3-ratelimitresponse).
 
-Setting _all_ the following environment variables to the header name to use:
-1. `LIMIT_LIMIT_HEADER` - The value will be the current limit value closest to being triggered, currently the optional quota policies are not added
-1. `LIMIT_REMAINING_HEADER` - The value will be the remaining quota
-1. `LIMIT_RESET_HEADER` - The value will be the number of seconds until the limit is being reset
+The following environment variables control the custom response feature:
+1. `LIMIT_RESPONSE_HEADERS_ENABLED` - Enables the custom response headers
+1. `LIMIT_LIMIT_HEADER` - The default value is "RateLimit-Limit", setting the environment variable will specify an alternative header name
+1. `LIMIT_REMAINING_HEADER` - The default value is "RateLimit-Remaining", setting the environment variable will specify an alternative header name
+1. `LIMIT_RESET_HEADER` - The default value is "RateLimit-Reset", setting the environment variable will specify an alternative header name
 
 # Contact
 
--- a/README.md
+++ b/README.md
@@ -627,7 +627,7 @@ When using multiple memcache nodes in `MEMCACHE_HOST_PORT=`, one should provide 
 to all ratelimiter instances to ensure that a particular cache key is always hashed to the same memcache node.
 
 # Custom headers
-Ratelimit service can be configured to return custom headers with the ratelimit information.
+Ratelimit service can be configured to return custom headers with the ratelimit information. It will populate the response_headers_to_add as part of the [RateLimitRespons](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/ratelimit/v3/rls.proto#service-ratelimit-v3-ratelimitresponse).
 
 Setting _all_ the following environment variables to the header name to use:
 1. `LIMIT_LIMIT_HEADER` - The value will be the current limit value closest to being triggered, currently the optional quota policies are not added
--- a/src/service/ratelimit.go
+++ b/src/service/ratelimit.go
@@ -163,7 +163,8 @@ func (this *service) shouldRateLimitWorker(
 
 	for i, descriptorStatus := range responseDescriptorStatuses {
 		// Keep track of the descriptor closest to hit the ratelimit
-		if descriptorStatus.CurrentLimit != nil &&
+		if this.customHeadersEnabled &&
+			descriptorStatus.CurrentLimit != nil &&
 			descriptorStatus.LimitRemaining < minLimitRemaining {
 			minimumDescriptor = descriptorStatus
 			minLimitRemaining = descriptorStatus.LimitRemaining
@@ -219,30 +220,10 @@ func (this *service) rateLimitRemainingHeader(descriptor *pb.RateLimitResponse_D
 
 func (this *service) rateLimitResetHeader(
 	descriptor *pb.RateLimitResponse_DescriptorStatus) *core.HeaderValue {
-	descriptor.DurationUntilReset.IsValid()
+
 	return &core.HeaderValue{
 		Key:   this.customHeaderResetHeader,
 		Value: strconv.FormatInt(utils.CalculateReset(&descriptor.CurrentLimit.Unit, this.customHeaderClock).GetSeconds(), 10),
-	}
-}
-
-func calculateReset(descriptor *pb.RateLimitResponse_DescriptorStatus, now int64) int64 {
-	sec := unitInSeconds(descriptor.CurrentLimit.Unit)
-	return sec - now%sec
-}
-
-func unitInSeconds(unit pb.RateLimitResponse_RateLimit_Unit) int64 {
-	switch unit {
-	case pb.RateLimitResponse_RateLimit_SECOND:
-		return 1
-	case pb.RateLimitResponse_RateLimit_MINUTE:
-		return 60
-	case pb.RateLimitResponse_RateLimit_HOUR:
-		return 60 * 60
-	case pb.RateLimitResponse_RateLimit_DAY:
-		return 60 * 60 * 24
-	default:
-		panic("unknown rate limit unit")
 	}
 }
 
--- a/src/service/ratelimit.go
+++ b/src/service/ratelimit.go
@@ -6,10 +6,10 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/envoyproxy/ratelimit/src/settings"
 	"github.com/envoyproxy/ratelimit/src/stats"
+	"github.com/envoyproxy/ratelimit/src/utils"
 
 	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	pb "github.com/envoyproxy/go-control-plane/envoy/service/ratelimit/v3"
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -625,6 +626,14 @@ descriptors will fail. Descriptors sent to Memcache should not contain whitespac
 When using multiple memcache nodes in `MEMCACHE_HOST_PORT=`, one should provide the identical list of memcache nodes
 to all ratelimiter instances to ensure that a particular cache key is always hashed to the same memcache node.
 
+# Custom headers
```
ysawa0, line 629: It'd be good to link to the official Envoy specs in this section so its more clear what is being set

src/service/ratelimit.go
```diff
@@ -132,7 +167,19 @@ func (this *service) shouldRateLimitWorker(
 	response := &pb.RateLimitResponse{}
 	response.Statuses = make([]*pb.RateLimitResponse_DescriptorStatus, len(request.Descriptors))
 	finalCode := pb.RateLimitResponse_OK
+
+	// Keep track of the descriptor which is closes to hit the ratelimit
```
ysawa0, line 171: nit typo: closest not closes

ysawa0:
Just so I'm understanding correctly, is this PR trying to implement the response of the X-RateLimit-Limit, X-RateLimit-Reset, X-RateLimit-Remaining info as specified in https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ratelimit/v3/rate_limit.proto ?

README.md
jespersoderlund: definitly, will add

jespersoderlund:
> Just so I'm understanding correctly, is this PR trying to implement the response of the X-RateLimit-Limit, X-RateLimit-Reset, X-RateLimit-Remaining info as specified in https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ratelimit/v3/rate_limit.proto ?

Yes, correct.

src/settings/settings.go
```diff
@@ -49,6 +49,14 @@ type Settings struct {
 	CacheKeyPrefix             string  `envconfig:"CACHE_KEY_PREFIX" default:""`
 	BackendType                string  `envconfig:"BACKEND_TYPE" default:"redis"`
 
+	// Settings for optional returning of custom headers
+	// value: the current limit
+	HeaderRatelimitLimit string `envconfig:"LIMIT_LIMIT_HEADER" default:""`
+	// value: remaining count
+	HeaderRatelimitRemaining string `envconfig:"LIMIT_REMAINING_HEADER" default:""`
+	// value: remaining seconds
+	HeaderRatelimitReset string `envconfig:"LIMIT_RESET_HEADER" default:""`
```
ysawa0, line 58: IMO I think it would be better to have an additional flag, something like `XRatelimitResponseHeadersEnabled` rather than having the toggle logic in the service struct and have these 3 settings have a default value corresponding to the header names specified in the RFC, eg: "X-RateLimit-Limit" "X-RateLimit-Remaining" and "X-RateLimit-Reset"

test/service/ratelimit_test.go
```diff
@@ -176,6 +187,99 @@ func TestService(test *testing.T) {
...
+				{Key: "A-Ratelimit-Limit", Value: "10"},
+				{Key: "A-Ratelimit-Remaining", Value: "0"},
+				{Key: "A-Ratelimit-Reset", Value: "58"},
+			},
+		},
+		response)
+	t.assert.Nil(err)
+
+	t.assert.EqualValues(2, t.statStore.NewCounter("config_load_success").Value())
+	t.assert.EqualValues(1, t.statStore.NewCounter("config_load_error").Value())
```
ysawa0, line 280: do we need to check for config loads in this test?

test/service/ratelimit_test.go
```diff
@@ -176,6 +187,99 @@ func TestService(test *testing.T) {
...
+
+	// Config load failure.
+	t.configLoader.EXPECT().Load(
+		[]config.RateLimitConfigToLoad{{"config.basic_config", "fake_yaml"}}, gomock.Any()).Do(
+		func([]config.RateLimitConfigToLoad, stats.Manager) {
+			defer barrier.signal()
+			panic(config.RateLimitConfigError("load error"))
+		})
+	t.runtimeUpdateCallback <- 1
+	barrier.wait()
```
ysawa0, line 250: do we need to check for config loads in this test?

test/service/ratelimit_test.go
```diff
@@ -176,6 +187,99 @@ func TestService(test *testing.T) {
...
+	defer t.controller.Finish()
+	service := t.setupBasicService()
+
+	// Config reload.
+	barrier := newBarrier()
+	t.configLoader.EXPECT().Load(
+		[]config.RateLimitConfigToLoad{{"config.basic_config", "fake_yaml"}}, gomock.Any()).Do(
+		func([]config.RateLimitConfigToLoad, stats.Manager) { barrier.signal() }).Return(t.config)
+	t.runtimeUpdateCallback <- 1
+	barrier.wait()
```
ysawa0, line 210: do we need to check for config loads in this test?

test/service/ratelimit_test.go
```diff
@@ -176,6 +187,99 @@ func TestService(test *testing.T) {
...
+		&pb.RateLimitResponse{
+			OverallCode: pb.RateLimitResponse_OVER_LIMIT,
+			Statuses: []*pb.RateLimitResponse_DescriptorStatus{
+				{Code: pb.RateLimitResponse_OVER_LIMIT, CurrentLimit: limits[0].Limit, LimitRemaining: 0},
+				{Code: pb.RateLimitResponse_OK, CurrentLimit: nil, LimitRemaining: 0},
+			},
+			ResponseHeadersToAdd: []*core.HeaderValue{
+				{Key: "A-Ratelimit-Limit", Value: "10"},
+				{Key: "A-Ratelimit-Remaining", Value: "0"},
+				{Key: "A-Ratelimit-Reset", Value: "58"},
```
ysawa0, line 236: I'm looking at the proto definitions and shouldn't the values like `{Key: "A-Ratelimit-Remaining", Value: "0"}` be set under `LimitRemaining` in pb.RateLimitResponse_DescriptorStatus ?

src/service/ratelimit.go
```diff
@@ -142,14 +189,71 @@ func (this *service) shouldRateLimitWorker(
...
 			}
 		}
 	}
 
+	// Add Headers if requested
+	if this.customHeadersEnabled && minimumDescriptor != nil {
+		response.ResponseHeadersToAdd = []*core.HeaderValue{
+			this.rateLimitLimitHeader(minimumDescriptor),
+			this.rateLimitRemainingHeader(minimumDescriptor),
+			this.rateLimitResetHeader(minimumDescriptor, this.customHeaderClock.Now().Unix()),
```
ysawa0, line 204: I believe you need to set these under `RateLimitResponse_DescriptorStatus` in the proto not ResponseHeadersToAdd

src/service/ratelimit.go
```diff
@@ -142,14 +189,71 @@ func (this *service) shouldRateLimitWorker(
...
+	descriptor *pb.RateLimitResponse_DescriptorStatus, now int64) *core.HeaderValue {
+
+	return &core.HeaderValue{
+		Key:   this.customHeaderResetHeader,
+		Value: strconv.FormatInt(calculateReset(descriptor, now), 10),
+	}
+}
+
+func calculateReset(descriptor *pb.RateLimitResponse_DescriptorStatus, now int64) int64 {
+	sec := unitInSeconds(descriptor.CurrentLimit.Unit)
```
ysawa0, line 238: Some comments here would be good for understanding how the response is calculated

src/settings/settings.go
jespersoderlund: Sure, I can go with that

test/service/ratelimit_test.go
jespersoderlund: Removed them from the test that I cloned from the existing one. TBH lots of the tests in the project are quite long and do a lot of different checks

src/service/ratelimit.go
jespersoderlund: I don't think so, when I run the current version from envoy, the custom headers shows up

test/service/ratelimit_test.go
jespersoderlund: But they are headers so it needs to go into the ResponseHeadersToAdd element

src/settings/settings.go
ysawa0: I guess what I'm trying to say is -- most likely 99% of people will use the official header names as stated in the RFC/Envoy docs so I think we should set those as the default values in here. That means we would need an 'enabled' flag in the settings to determine if these responses are enabled rather than enabling them if the strings are not empty. I think that would be cleaner as well.

src/service/ratelimit.go
ysawa0: got it, makes sense now. would we want to update the DescriptorStatus field with the same info as well for completeness?
```
type RateLimitResponse_DescriptorStatus struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	// The response code for an individual descriptor.
	Code RateLimitResponse_Code `protobuf:"varint,1,opt,name=code,proto3,enum=envoy.service.ratelimit.v3.RateLimitResponse_Code" json:"code,omitempty"`
	// The current limit as configured by the server. Useful for debugging, etc.
	CurrentLimit *RateLimitResponse_RateLimit `protobuf:"bytes,2,opt,name=current_limit,json=currentLimit,proto3" json:"current_limit,omitempty"`
	// The limit remaining in the current time unit.
	LimitRemaining uint32 `protobuf:"varint,3,opt,name=limit_remaining,json=limitRemaining,proto3" json:"limit_remaining,omitempty"`
	// Duration until reset of the current limit window.
	DurationUntilReset *duration.Duration `protobuf:"bytes,4,opt,name=duration_until_reset,json=durationUntilReset,proto3" json:"duration_until_reset,omitempty"`
}
```

src/settings/settings.go
jespersoderlund: Yes, I agree I'll do that change with an "Enable" flag, but the default values to be identical to the ones that are generated by default by Envoy might be a bit unecessary. In which case enabling the custom headers, will generate the same headers as Envoy does by default. I can do that if you think it makes sense, but perhaps it would make more sense to use the proper-headers from the draft spec, ie. without the "X-" (https://datatracker.ietf.org/doc/html/draft-polli-ratelimit-headers-03)

src/service/ratelimit.go
jespersoderlund: I don't understand, these are already being set by the current implementation and doesn't have anything todo with the custom headers

src/settings/settings.go
jespersoderlund: Added the draft standard headers as the default values

src/service/ratelimit.go
```diff
@@ -132,7 +156,19 @@ func (this *service) shouldRateLimitWorker(
...
+	minLimitRemaining := MaxUint32
+	var minimumDescriptor *pb.RateLimitResponse_DescriptorStatus = nil
+
 	for i, descriptorStatus := range responseDescriptorStatuses {
+		// Keep track of the descriptor closest to hit the ratelimit
+		if descriptorStatus.CurrentLimit != nil &&
+			descriptorStatus.LimitRemaining < minLimitRemaining {
+			minimumDescriptor = descriptorStatus
+			minLimitRemaining = descriptorStatus.LimitRemaining
+		}
```
ysawa0, line 170: do we just want to gate this logic by checking `this.customHeadersEnabled` in the conditional as well? doubt it has much performance implications but may be good just to be sure

src/service/ratelimit.go
```diff
@@ -142,14 +178,74 @@ func (this *service) shouldRateLimitWorker(
...
+	// How much of the limit is remaining
+	return &core.HeaderValue{
+		Key:   this.customHeaderRemainingHeader,
+		Value: strconv.FormatUint(uint64(descriptor.LimitRemaining), 10),
+	}
+}
+
+func (this *service) rateLimitResetHeader(
+	descriptor *pb.RateLimitResponse_DescriptorStatus) *core.HeaderValue {
+	descriptor.DurationUntilReset.IsValid()
```
ysawa0, line 222: this line looks like a noop?

src/service/ratelimit.go
```diff
@@ -142,14 +178,74 @@ func (this *service) shouldRateLimitWorker(
...
+	return &core.HeaderValue{
+		Key:   this.customHeaderResetHeader,
+		Value: strconv.FormatInt(utils.CalculateReset(&descriptor.CurrentLimit.Unit, this.customHeaderClock).GetSeconds(), 10),
+	}
+}
+
+func calculateReset(descriptor *pb.RateLimitResponse_DescriptorStatus, now int64) int64 {
```
ysawa0, line 229: this func doesn't seem to be used

src/service/ratelimit.go
```diff
@@ -142,14 +178,74 @@ func (this *service) shouldRateLimitWorker(
...
+	sec := unitInSeconds(descriptor.CurrentLimit.Unit)
+	return sec - now%sec
+}
+
+func unitInSeconds(unit pb.RateLimitResponse_RateLimit_Unit) int64 {
```
ysawa0, line 234: same here

src/service/ratelimit.go
ysawa0: Got it. Finally had some time to test the code in more detail and makes sense now.

README.md
```diff
@@ -625,6 +626,14 @@ descriptors will fail. Descriptors sent to Memcache should not contain whitespac
...
+Ratelimit service can be configured to return custom headers with the ratelimit information. It will populate the response_headers_to_add as part of the [RateLimitRespons](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/ratelimit/v3/rls.proto#service-ratelimit-v3-ratelimitresponse).
```
ysawa0, line 630: typo on `RateLimitRespons`

README.md
```diff
@@ -625,6 +626,14 @@ descriptors will fail. Descriptors sent to Memcache should not contain whitespac
...
+
+Setting _all_ the following environment variables to the header name to use:
+1. `LIMIT_LIMIT_HEADER` - The value will be the current limit value closest to being triggered, currently the optional quota policies are not added
+1. `LIMIT_REMAINING_HEADER` - The value will be the remaining quota
+1. `LIMIT_RESET_HEADER` - The value will be the number of seconds until the limit is being reset
+
```
ysawa0, line 636: this section needs to be updated to reflect the new ENABLED flag
/wait

src/service/ratelimit.go
jespersoderlund: Yes, they got left after I discovered existing time-utils and refactored to use those, but the these got left

ysawa0 (approved):

ysawa0:
looks great, thank you!

Issue #293, opened 2021-09-21
Proposal: Shadow mode


jespersoderlund:
It would be very helpful to roll-out rate limiting on existing systems with the help of a shadow-mode where the logic and statistics are run exactly as if the limits would have been in place, but where the response is always OK.

It has been added in a fork by https://github.com/replicon/ratelimit/pull/12 by @repl-david-winiarski and would be great to have something like this  integrated into the upstream.

The solution selected by replicon is to have a `shadowmode: true` attribute in the rate limiting definitions so that it can be selectively applied.

mattklein123:
Sure makes sense to me.

jespersoderlund:
Working on a PR inspired by replicon's fork. It's a bit intertwined with additional changes so but I'll push for review in a day or two hopefully

jespersoderlund:
Thanks

PR #294, opened 2021-09-26
293 shadowmode

jespersoderlund:
Addressing issue: #293 

Adding shadow-mode:
- Rule based shadow mode so that specific keys can be put in shadow mode
- Global shadow-mode where all rate limit requests will be OK

Statistics on key-based and global shadow mode


ysawa0:
Looks like CI is failing, can you take a look
/wait

README.md
```diff
@@ -430,6 +478,19 @@ go run main.go -domain test \
...
+
+There is a global shadow-mode which can make it easier to introduce rate limiting into an existing service landscape. It will override whatever result is returned by the regular rate limiting process.
+
+## Configuration
+The global shadow mode is configured with an environment variable
+
+Setting environment variable`SHADOW_MODE` to `true` will enable the feature.
+
+## Statistics
+There is an additional service-level statistics generated that will increment whenever the global shadow mode has overridden a rate limiting result.
```
ysawa0, line 491: What is the difference between the `over_limit.shadow_mode` and `over_limit.over_limit` metrics? If I'm understanding it correctly, if shadow mode is on for a rate limit, these 2 metrics would always equal. Does it exist so we can observe the shadow testing easier?

jespersoderlund: The reason for having the shadow_mode metrics for global is to be able to distinguish the situation in the metrics series if the configuration is changed over time and that it's not a regular "over_limit". In the global shadow-mode I would like to affect as little as possible "inside" of the service, but want to be able to identify in the metric series that it was not actually ratelimiting due to shadow_mode

jespersoderlund: I can add it to the docker-compose example to make it easy to test

jespersoderlund: Added shadow_mode to the example docker-compose and docs

docker-compose-example.yml
```diff
@@ -26,7 +26,9 @@ services:
       - ./examples/prom-statsd-exporter/conf.yaml:/etc/statsd-exporter/conf.yaml
 
   ratelimit:
-    image: envoyproxy/ratelimit:master
+    build:
+      context: .
+      dockerfile: Dockerfile
```
ysawa0, line 31: while it makes sense to have a test env that uses latest local code, I think we should make another docker-compose file for that. once we merged this PR in, the `master` image should pick up all shadowmade changes. 

kind of separate but I was thinking if we had a docker compose env that builds off the latest code, we could use it to do basic integration testing + performance testing in GH Actions via curl and timing the response + checking status codes.

README.md
ysawa0: lgtm, thanks!

ysawa0 (commented):
took a quick level look, the changes make sense to me and I think shadow mode is an option that is definitely needed. left some comments, will take a deeper look later.

docker-compose-example.yml
jespersoderlund: I totally agree, that's what we typically do at my place. I'll put together an integration test and push

ysawa0:
Took a 2nd look and everything else lgtm, so just pending some changes and merge conflicts.

test/service/ratelimit_test.go
```diff
@@ -76,7 +78,7 @@ func commonSetup(t *testing.T) rateLimitServiceTestSuite {
 	return ret
 }
 
-func (this *rateLimitServiceTestSuite) setupBasicService() ratelimit.RateLimitServiceServer {
+func (this *rateLimitServiceTestSuite) setupBasicService(shadow_mode bool) ratelimit.RateLimitServiceServer {
```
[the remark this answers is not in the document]
jespersoderlund: leftover from some earlier code-structure during dev.

docker-compose-example.yml
jespersoderlund: I added a self-contained docker-compose with some simple end-to-end-tests.

Make file target will trigger the tests and exit with a non-0 code if any fails

ysawa0 (approved):

ysawa0:
Looks good, thanks!

PR #295, opened 2021-09-27
Fix MEMCACHED_SRV support

petedmarsh:
This did not work in practice due to a mix up of

    func example(serverList memcache.ServerList)

and

    func example(serverList *memcache.ServerList)

The code used the first case and did not correctly update a given
memcache.ServerList instance, instead a new one was created, updated
then forgot.

### DIFF
--- a/src/memcached/cache_impl.go
+++ b/src/memcached/cache_impl.go
@@ -175,7 +175,7 @@ func (this *rateLimitMemcacheImpl) Flush() {
 	this.waitGroup.Wait()
 }
 
-func refreshServersPeriodically(serverList memcache.ServerList, srv string, d time.Duration, finish <-chan struct{}) {
+func refreshServersPeriodically(serverList *memcache.ServerList, srv string, d time.Duration, finish <-chan struct{}) {
 	t := time.NewTicker(d)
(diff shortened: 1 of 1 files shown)


nezdolik (commented):
this should have ideally been covered by testing

mattklein123:
> this should have ideally been covered by testing

Agreed. Is there any kind of test we can add to catch this?

petedmarsh:
@nezdolik @mattklein123 I do completely agree with you, this code path is tricky to test as it does DNS lookups and I'm really not sure how to go about writing tests for it (I have fairly limited go experience).

I guess mocking would be the way, if you have any pointers or links that would be useful I can go away and try to add some. 

This is currently running in production and does fix the broken code that's currently in main so perhaps it is worth merging? But, genuinely I am happy to write some tests I just need some help :)

mattklein123:
We do have mocking support already used in this library, but I'm not sure exactly the best way to refactor this code to make it mockable. I'm OK merging this for now but if you can try to follow up with testing that would be appreciated. cc @ysawa0 who might have better ideas.

mattklein123 (approved):

PR #298, opened 2021-09-30
Adding tests requested in #295 

petedmarsh:
I recent raised #295 to fix a bug, this is a bug that could have been caught with tests but no tests had been written (I wrote the original code, so it is my fault).

I've had a stab here at writing some tests for this case, they required a small amount of refactoring. I fully expect that I can improve upon them and I am happy to do so. For one thing the test file is in the main `src` tree rather than `test` as it seemed silly to me to make these functions public, but it also seems silly to have a test here.

Anyway, I am happy to improve and add more, I am raising early in the hope I can get some feedback (though I appreciate that that isn't your job to do so!).


petedmarsh:
This'll need to be rebased anyway as 361e3ed is in main under a different SHA, I will sort that out.

mattklein123:
cc @nezdolik @ysawa0 for comments.

nezdolik:
Thanks, I will review this week.

src/srv/srv.go
```diff
@@ -21,7 +27,7 @@ func ParseSrv(srv string) (string, string, string, error) {
 	return matches[1], matches[2], matches[3], nil
 }
 
-func ServerStringsFromSrv(srv string) ([]string, error) {
+func (s DnsSrvResolver) ServerStringsFromSrv(srv string) ([]string, error) {
```
nezdolik, line 30: nit: use more descriptive var name instead of `s`

src/memcached/cache_impl_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+	serverList.Each(func(addr net.Addr) error {
+		actualMemcacheHosts = append(actualMemcacheHosts, addr.String())
+		return nil
+	})
+
+	assert.Equal([]string{mockMemcacheHostPort}, actualMemcacheHosts)
+}
+
+func TestRefreshServerSetsServersOnNonEmptyServerList(t *testing.T) {
```
nezdolik, line 41: nit: i think more descriptive name this test case would be `...OverridesServersOnNonEmptyServerList`

nezdolik:
Thank you for adding the tests! Few nits from my side, and looks like commits are missing sign-off.

petedmarsh:
@nezdolik I have updated everything as per your comments

src/srv/srv.go
nezdolik: @petedmarsh ping on this

nezdolik:
@petedmarsh looks like one more nit left and then you can mark this PR as ready for review (instead of Draft)

nezdolik (approved):

mattklein123 (approved):
Awesome, thanks!

PR #305, opened 2021-10-22
Fail healthcheck when no config files loaded

fishcakez:
To improve ready handling of the healthcheck, await a config to be loaded before
passing the health check or if there are no config files to load. Once a config
file is loaded pass the healthcheck.

Add a Pass() method to the HealthChecker and start it in a failed state until
a config file is successfully loaded.

Signed-off-by: James Fish <jfish@pinterest.com>

### DIFF
--- a/src/server/health.go
+++ b/src/server/health.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"sync"
 	"sync/atomic"
 	"syscall"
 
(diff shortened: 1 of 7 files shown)


src/server/health.go
```diff
@@ -12,24 +13,30 @@ import (
 )
 
 type HealthChecker struct {
-	grpc *health.Server
-	ok   uint32
-	name string
+	sync.Mutex
```
ysawa0, line 16: make private variable `mu sync.Mutex`

ysawa0:
@fishcakez while this change mostly makes sense to me, just curious was there a certain problem you were running into that this PR is trying to solve? also there are some merge conflicts you will need to fix.

fishcakez:
@ysawa0 we are migrating to a new control plane system used to deliver the config files to disk for ratelimit service. It's possible to more gracefully handle a control plane failure if the healthcheck can signal no configuration loaded to avoid routing to that host. If the deployment system (of the ratelimit service itself) can deploy the ratelimit service on a new host without blocking on the control plane to have loaded all configs then its possible to recover on that host once the control plane is healthy and has delivered the config to disk. This is possible because the ratelimit service got started and becomes healthy (and routable) once the file is on disk.

Note we currently deploy ratelimit service with the sidecar model and not its own service/cluster. Therefore scaling can occur regularly and we want to be able to automatically recover without interfering with the main service.

tinder-yukisawa:
@fishcakez That makes sense. Happy to finish review once merge conflicts are fixed.

PR #314, opened 2021-12-17
Support for "Memory" backend.

konrad-kaminski:
This is the implementation of pure memory-based backend. It is useful in situations where we do not want to have a (heavier) memcached/redis backend, but still want to use global rate limiting functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,7 +392,7 @@ descriptors:
       requests_per_unit: 100
 ```
 
-For an unlimited descriptor, the request will not be sent to the underlying cache (Redis/Memcached), but will be quickly returned locally by the ratelimit instance.
+For an unlimited descriptor, the request will not be sent to the underlying cache (Redis/Memcached/Memory), but will be quickly returned locally by the ratelimit instance.
 This can be useful for collecting statistics, or if one wants to define a descriptor that has no limit but the client wants to distinguish between such descriptor and one that does not exist.
 
 The return value for unlimited descriptors will be an OK status code with the LimitRemaining field set to MaxUint32 value.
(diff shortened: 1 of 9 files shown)


ysawa0:
Hi @konrad-kaminski, this looks like it could be useful but I'm not sure how it would be used. Would memory mode be used only in the situation where there is only a single instance of ratelimit running in your infrastructure and all clients call that single instance? Otherwise the rate limits would not be shared globally.

konrad-kaminski:
Yes, that's exactly the scenario we want to support i.e. a single instance of ratelimit would support many clients to achieve global rate limits.

ysawa0:
I see. I understand the use case but I'm not sure if that's a design we should support as it's not exactly following best practices (Not HA and if the single ratelimit instance dies, all limits are lost). cc @mattklein123 I wonder if you have any thoughts on this?

mattklein123:
> cc @mattklein123 I wonder if you have any thoughts on this?

Yeah I had the same thought about a single point of failure and HA. It doesn't seem like this is how this service should be deployed?

konrad-kaminski:
I understand that use cases are limited. We mostly want to support a per-second limits, thus HA would be nice, but is not necessary.

konrad-kaminski:
@mattklein123  , @ysawa0 What are your final thoughts? We're kind of stuck waiting for the decision.

mattklein123:
My general feeling is this isn't something we should support since there is no possible HA solution. @ysawa0 any objection?

ysawa0:
> My general feeling is this isn't something we should support since there is no possible HA solution. @ysawa0 any objection?

Yes, I agree with that sentiment.

konrad-kaminski:
OK. I'm closing this PR. For those who'd like that kind of functionality, we've created a fork at https://github.com/allegro/ratelimit/.

PR #316, opened 2021-12-19
Fix memcached shards being inconsistent when looking up hosts by SRV

petedmarsh:
SRV records return hosts in an arbitrary order, but the memcached
library in use relies on the order of the hosts for sharding (i.e.
the i-th host passed to the client will host shard i). If multiple
ratelimitinstances looked up the same set of memcached hosts via
SRV they could recieve them in different orders and begin sharding
descriptor keys to separate hosts, leading to non-global rate
limiting (as the sum of hits would be split accross hosts, meaning
each ratelimit instance would have an incomplete total of hits).

This fixes the issue by sorting the hosts returned by the SRV
lexicographically. Both weight and priority are ignored; weight
was not previously handled and priority is not possible to handle
(as the memcached library cannot take a priority into account when
determining shards).

### DIFF
--- a/src/srv/srv.go
+++ b/src/srv/srv.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"regexp"
+	"sort"
 
 	logger "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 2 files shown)


ysawa0 (approved):
Lgtm, thanks for the fix!

Issue #326, opened 2022-04-12
Support Tracing


ethernoy:
I am experiencing some difficulty when troubleshooting a slow ratelimit installation. I believe adding tracing of any format (preferrably zipkin or openTelemetry) to the ratelimit service can be of great help.

Happy to submit a PR if this is ok.

mattklein123:
Sure, though I would go straight to Open Telemetry.

ethernoy:
> Sure, though I would go straight to Open Telemetry.

Great, then let me work on it

ethernoy:
> Sure, though I would go straight to Open Telemetry.

@mattklein123 Have a question in mind: if we are going to send out OTEL span when calling memcached and redis, it probably needs the caller to pass the parent traceId in gRPC request body, which isn't around for now according to the `proto` file in envoyproxy project: 

https://github.com/envoyproxy/envoy/blob/main/api/envoy/service/ratelimit/v2/rls.proto
https://github.com/envoyproxy/envoy/blob/main/api/envoy/service/ratelimit/v3/rls.proto

Should we consider to add such data in the ratelimit request protocol? Also we will need to add the implementation in envoyproxy in order to produce a full trace.

ethernoy:
> Yeah we can add this to the API for sure.

Nice, then let me finish the implementation from ratelimit side first

zirain:
fixed by https://github.com/envoyproxy/ratelimit/pull/332？

ChezCrawford:
Can we close this if it was taken care of by https://github.com/envoyproxy/ratelimit/pull/332 ?

PR #332, opened 2022-04-27
Implemented OpenTelemetry Instrumentation

ethernoy:
Signed-off-by: Ethern Su <ehaprime@gmail.com>

Related issue: https://github.com/envoyproxy/ratelimit/issues/326

I instrumented ratelimit with the [otelgrpc](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/google.golang.org/grpc/otelgrpc) project and added necessary span during redis pipeline execution and memcached fetch call. Documentation added to readme.md.

Test environment topology:

```
┌────────────────┐                                              browser
│ratelimit client│                                                 │
└───────┬────────┘                                                 │ http:16686
        │                                                          │
  ┌─────▼─────┐    http:4318    ┌─────────────┐   http:14250   ┌───▼──┐
  │ ratelimit ├────────────────►│OpenTelemetry├───────────────►│Jaeger│
  └─────┬─────┘    OTLP span    │  Collector  │   Jaeger span  └──────┘
        │                       └─────────────┘
┌───────▼───────┐               convert + batch
│redis/memcached│
└───────────────┘
```

Screenshot of the working tracing of both redis-backed and memcached-backed ratelimit service using Jaeger as visualization tool:

Memcached
image

Redis
image

### DIFF
--- a/src/service_cmd/runner/runner.go
+++ b/src/service_cmd/runner/runner.go
@@ -78,7 +78,7 @@ func createLimiter(srv server.Server, s settings.Settings, localCache *freecache
 func (runner *Runner) Run() {
 	s := runner.settings
 	if s.TracingEnabled {
-		tp := trace.Init(s.TracingExporterProtocol, s.TracingServiceName, s.TracingServiceNamespace, s.TracingServiceInstanceId)
+		tp := trace.InitProductionTraceProvider(s.TracingExporterProtocol, s.TracingServiceName, s.TracingServiceNamespace, s.TracingServiceInstanceId)
 		defer func() {
 			if err := tp.Shutdown(context.Background()); err != nil {
 				logger.Printf("Error shutting down tracer provider: %v", err)
--- a/src/trace/trace.go
+++ b/src/trace/trace.go
@@ -22,7 +22,7 @@ var (
 	testSpanExporterMu sync.Mutex
 )
 
-func Init(protocol string, serviceName string, serviceNamespace string, serviceInstanceId string) *sdktrace.TracerProvider {
+func InitProductionTraceProvider(protocol string, serviceName string, serviceNamespace string, serviceInstanceId string) *sdktrace.TracerProvider {
 	client := createClient(protocol)
 	exporter, err := otlptrace.New(context.Background(), client)
 	if err != nil {


ethernoy:
@mattklein123 Hi, can you help reviewing this PR?

src/service/ratelimit.go
```diff
@@ -240,6 +245,15 @@ func (this *service) ShouldRateLimit(
 	ctx context.Context,
 	request *pb.RateLimitRequest) (finalResponse *pb.RateLimitResponse, finalError error) {
 
+	// Generate trace
+	_, span := tracer.Start(ctx, "ShouldRateLimit Execution",
```
mattklein123, line 249: What is the cost of all of these pan creations if tracing is disabled?

src/service_cmd/runner/runner.go
```diff
@@ -75,6 +77,16 @@ func createLimiter(srv server.Server, s settings.Settings, localCache *freecache
 
 func (runner *Runner) Run() {
 	s := runner.settings
+	if s.TracingEnabled {
```
mattklein123, line 80: I want to make sure I understand how this is implemented: Right now all of these traces are independent, right? Because we aren't passing any incoming span data, right? Are you still planning on changing the API to optionally pass on context?

mattklein123, line 80: I want to make sure I understand how this is implemented: Right now all of these traces are independent, right? Because we aren't passing any incoming span data, right? Are you still planning on changing the API to optionally pass on context?

src/server/server_impl.go
```diff
@@ -88,6 +96,14 @@ func NewJsonHandler(svc pb.RateLimitServiceServer) func(http.ResponseWriter, *ht
 			return
 		}
 
+		// Generate trace
+		_, span := tracer.Start(context.TODO(), "NewJsonHandler Remaining Execution",
```
mattklein123, line 100: Without a real context is there any real use for this?

mattklein123, line 100: Without a real context is there any real use for this?

mattklein123 (changes requested):
Thanks for adding this. A few questions to get started.

Also, is it possible to add any kind of testing for this? Even if mocked? It would be nice to make sure we are creating the spans where we think we should be.

src/service/ratelimit.go
ethernoy: When tracing is diabled, no span processor or exporter is added to the trace provider, hence tracer.Start and tracer.End are basically doing noop.

src/server/server_impl.go
ethernoy: I have confusion on this part too. The NewJsonHandler for http endpoint calls the `ShouldRateLimit` without passing a context. Should we initiate a background context in the handler and pass it to both `ShouldRateLimit` function and the tracer for correctly handling the context?

ethernoy: Resolving this conversation as it is a duplicate.

src/service_cmd/runner/runner.go
ethernoy: Right now all spans in one call to `ShouldRateLimit` gRPC function are associated in one trace using the context (e.g. gRPC client span + gRPC server span + redis pipeline execution span, you can refer to screenshot above). But in the real use case where it is the envoyproxy calling the ratelimit, the envoyproxy needs to instrument the gRPC call to ratelimit service in order to associate the trace.

ethernoy: Resolving this conversation as it is a duplicate.

ethernoy:
> Thanks for adding this. A few questions to get started.
> 
> Also, is it possible to add any kind of testing for this? Even if mocked? It would be nice to make sure we are creating the spans where we think we should be.

Yes we can test it using a docker-based otlp-collector for trace transformation and a all-in-one jaeger for trace storage and visualization. Commands attached:

```bash
# this command starts a otlp-collector that transforms otlp trace to jaeger trace. Predefined configuration is already in example directory.
docker run --name otlp -d -p 4318 -p 4317 -v examples/otlp-collector:/tmp/otlp-collector otel/opentelemetry-collector:0.48.0 -- --config /tmp/otlp-collector/config.yaml
otelcol-contrib --config examples/otlp-collector/config.yaml
```

```bash
# this command starts a jaeger all-in-one application to store trace in memory and perform visualization on port :16686
docker run -d --name jaeger -p 16686:16686 -p 14250:14250 jaegertracing/all-in-one:1.33
```

ethernoy:
I have tested running an Istio + envoyproxy service mesh with ratelimit service backed by the OTLP instrumented ratelimit docker image, where the envoyproxy uses Zipkin when sending spans while the ratelimit service uses OTLP. End to end test indicates that Jaeger is not able to associate the spans sent from envoyproxy and the spans sent from ratelimit service:

image

image

Note both traces are actually produced by a single API call on the Istio ingress gateway. Apparently, when ratelimit tracer got the request context, it cannot find a valid trace ID in the OTLP protocol, hence it generates a new trace chain. This should be resolvable by using OTLP to instrument envoyproxy.

I found there is already a PR related to this issue in envoyproxy project
https://github.com/envoyproxy/envoy/pull/20281

ethernoy:
> Thanks this makes sense. What I mean by testing is can you add some automated tests that make sure we make spans that have the right info in them. Is there some type of mock tracing provider that can be used for unit / local integration tests?

Sure, let me add it later.

ethernoy:
> Thanks this makes sense. What I mean by testing is can you add some automated tests that make sure we make spans that have the right info in them. Is there some type of mock tracing provider that can be used for unit / local integration tests?

Can you help to identify the appropriate places to record in tracing? Right now I only instrumented the following function calls:

- all gRPC call to the server
- shouldRateLimit function
- Redis pipeline execution
- Memcached query execution

mattklein123:
> Any suggestion on how to correctly test this?

I would recommend just creating a mock interface for creating spans, and then using that in tests and having the real implementation call the real APIs. It shouldn't be too hard I don't think.

I think the spans you have to start with sgtm.

ethernoy:
> > Any suggestion on how to correctly test this?
> 
> I would recommend just creating a mock interface for creating spans, and then using that in tests and having the real implementation call the real APIs. It shouldn't be too hard I don't think.
> 
> I think the spans you have to start with sgtm.

Just implemented the trace unit test on ratelimit, memcached impl and redis impl.

mattklein123:
Please don't force push once you open a PR, it makes reviews much more difficult. Also you need to fix DCO. At this point feel free to squash it all and force push to fix DCO, thank you.

ethernoy:
> Please don't force push once you open a PR, it makes reviews much more difficult. Also you need to fix DCO. At this point feel free to squash it all and force push to fix DCO, thank you.

gotcha, repushed.

test/memcached/cache_impl_test.go
```diff
@@ -648,6 +654,48 @@ func TestNewRateLimitCacheImplFromSettingsWhenHostAndPortAndSrvAreBothSet(t *tes
 	})
 }
 
+func TestMemcachedTracer(t *testing.T) {
```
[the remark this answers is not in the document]
ethernoy: sure, I suggest placing the spanExporter initialization code to a shared helper. Any recommendation on which path to put the shared helper package?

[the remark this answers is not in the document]
mattklein123: I don't have a strong opinion on where to put it. Feel free to make a new file if you like.

PR #334, opened 2022-05-08
reduce dependency for docker-compose

zackzhangkai:
Signed-off-by: zackzhangkai <zackzhangkai@gmail.com>

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,15 +9,6 @@ services:
     networks:
       - ratelimit-network
 
-  memcached:
-    image: memcached:alpine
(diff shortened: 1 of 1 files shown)


ysawa0:
Hi @zackzhangkai thanks for the PR. Could you explain what the benefit of the changes are? afaik the memcached container is used to test the memcached backend implementation?

zackzhangkai:
@ysawa0 Got it. I used to think we should reduce dependencies to make it lighter. Once I saw the docker-compose file, I got confused, for memcache and redis take the same function.

Issue #339, opened 2022-06-20
The docker-compose example is not starting envoy proxy properly


deferraz:
Hi! Looks like the latest changes in the project are not up-to-date with the envoy configuration example

Tried to run `docker-compose -f docker-compose-example.yml up --build --remove-orphans`

but the envoy-proxy and envoy-proxy-mock are not starting correctly, with the same error:
```
envoy-proxy_1  | [2022-06-20 17:50:29.237][1][info][config] [source/server/configuration_impl.cc:101] loading 1 listener(s)
envoy-proxy_1  | [2022-06-20 17:50:29.240][1][critical][main] [source/server/server.cc:117] error initializing configuration '/etc/envoy/envoy.yaml': Didn't find a registered implementation for 'envoy.filters.http.router' with type URL: ''
envoy-mock_1   | [2022-06-20 17:50:29.193][1][info][admin] [source/server/admin/admin.cc:134] admin address: 0.0.0.0:8001
ratelimit_1    | time="2022-06-20T17:50:29Z" level=debug msg="loading domain: mongo_cps"
ratelimit_1    | time="2022-06-20T17:50:29Z" level=debug msg="Creating stats for key: 'mongo_cps.database_users'"
envoy-mock_1   | [2022-06-20 17:50:29.194][1][info][config] [source/server/configuration_impl.cc:131] loading tracing configuration
envoy-mock_1   | [2022-06-20 17:50:29.194][1][info][config] [source/server/configuration_impl.cc:91] loading 0 static secret(s)
envoy-mock_1   | [2022-06-20 17:50:29.194][1][info][config] [source/server/configuration_impl.cc:97] loading 0 cluster(s)
ratelimit_1    | time="2022-06-20T17:50:29Z" level=debug msg="loading descriptor: key=mongo_cps.database_users ratelimit={requests_per_unit=500, unit=SECOND, unlimited=false, shadow_mode=false}"
ratelimit_1    | time="2022-06-20T17:50:29Z" level=debug msg="Creating stats for key: 'mongo_cps.database_default'"
ratelimit_1    | time="2022-06-20T17:50:29Z" level=debug msg="loading descriptor: key=mongo_cps.database_default ratelimit={requests_per_unit=500, unit=SECOND, unlimited=false, shadow_mode=false}"
envoy-mock_1   | [2022-06-20 17:50:29.194][1][info][config] [source/server/configuration_impl.cc:101] loading 1 listener(s)
[...]

PR #340, opened 2022-06-20
fix typed_config in example file for envoy.filters.http.router

deferraz:
Based on the Issue https://github.com/envoyproxy/ratelimit/issues/339 that I've opened, I think the envoy.filters.http.router http_filter was missing the typed_config for it , failing to startup during the compose up of the example of this project, this PR includes the type information so the envoy can start

Closes https://github.com/envoyproxy/ratelimit/issues/339

### DIFF
--- a/examples/envoy/mock.yaml
+++ b/examples/envoy/mock.yaml
@@ -26,7 +26,8 @@ static_resources:
                               inline_string: "Hello World"
                 http_filters:
                   - name: envoy.filters.http.router
-                    typed_config: {}
+                    typed_config:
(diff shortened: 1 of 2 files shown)


deferraz:
If any @envoyproxy/ratelimit-maintainers can review it, thanks in advance

mattklein123:
Sorry I don't have any quick ideas. cc @ysawa0 who might know?

ysawa0:
@deferraz it worked now when I re-ran it. may have been transient.

ysawa0 (approved):
thanks!

PR #209, opened 2021-01-27
Add implementation for hot reload of configuration with  same domain from multiple files.

anannaya:
Basically we enable the global ratelimit at the istio ingress layer , where the EnvoyFilter configuartion is applied with a generic domain name .  The current ratelimit service does not support updating/adding descriptors under same domain name.(Duplicate domain not supported) .  This implementation is about supporting the addition of descriptors existing domain. 

For Example:   Global gateway filter would be like below  with a generic domain name **vhost-ratelimit**
```
piVersion: networking.istio.io/v1alpha3
kind: EnvoyFilter
metadata:
  name: gateway-filter-ratelimit
  namespace: istio-system
spec:
  workloadSelector:
    labels:
      istio: ingressgateway-ratelimit
  configPatches:
    - applyTo: HTTP_FILTER
      match:
        context: GATEWAY
        listener:
          filterChain:
            filter:
              name: "envoy.http_connection_manager"
              subFilter:
                name: "envoy.router"
      patch:
        operation: INSERT_BEFORE
        value:
          name: envoy.rate_limit
          config:
            domain: vhost-ratelimit
            rate_limit_service:
              grpc_service:
                envoy_grpc:
                  cluster_name: rate_limit_service
                timeout: 0.25s
    - applyTo: CLUSTER
      match:
        cluster:
          service: ratelimit.ratelimit.svc.cluster.local
      patch:
        operation: ADD
        value:
          name: rate_limit_service
          type: STRICT_DNS
          connect_timeout: 0.25s
          lb_policy: ROUND_ROBIN
          http2_protocol_options: {}
          hosts:
            - socket_address:
                address: ratelimit.ratelimit.svc.cluster.local
                port_value: 8081

```
Next 
Basicaly I use a **kiwigrid/k8s-sidecar**. to reload the ratelimit  configuration from configmaps which is created by multiple applications. Sample Configmap looks like below , And which also has the Virtualservice specific envoyfilter which takes care of ratelimting the application requests.
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: configreload1
  labels:
    app: ratelimit
    ratelimit_config: "yes"
  namespace: pf
data:
  # statsd-mapping file
  configreload.yaml: |
    domain: vhost-ratelimit
    descriptors:
      - key: plann
        value: BASICCCCC
        descriptors:
          - key: accountttttt
            rate_limit:
              unit: minute
              requests_per_unit: 6
      - key: PATH
        value: "/productpage"
        rate_limit:
[...]

PR #357, opened 2022-08-24
Allow merge of configs referencing the same domain

### LINKED ISSUE #206: hot reload with a same domain is not supported.
l am using `RUNTIME_WATCH_ROOT` using hot reload the configuration. Is it possible to update the same domain name ?

level=error msg="error loading new configuration from runtime: config-1.yaml: config file cannot have empty domain"
time="2021-01-15T11:30:13Z" level=error msg="error loading new configuration from runtime: config.yaml: duplicate domain 'vhost-ratelimit' in config file"
time="2021-01-15T11:30:13Z" level=error msg="error loading new configuration from runtime: config.yaml: duplicate domain 'vhost-ratelimit' in config file"
time="2021-01-15T11:30:13Z" level=error msg="error loading new configuration from runtime: config.yaml: duplicate domain 'vhost-ratelimit' in config file"
time="2021-01-15T11:31:22Z" level=error msg="error loading new configuration from runtime: config-1.yaml: config file cannot have empty domain"

m-rcl:
Add MERGE_DOMAIN_CONFIG variable defaulting to false and mapping to the former behavior.
Setting to true will allow multiple files containing non duplicate keys and referencing the same domain to be merged.

* Fixes #206
* Continues #209

### DIFF
--- a/src/config_check_cmd/main.go
+++ b/src/config_check_cmd/main.go
@@ -15,7 +15,7 @@ import (
 	"github.com/envoyproxy/ratelimit/src/config"
 )
 
-func loadConfigs(allConfigs []config.RateLimitConfigToLoad) {
+func loadConfigs(allConfigs []config.RateLimitConfigToLoad, mergeDomainConfigs bool) {
 	defer func() {
 		err := recover()
 		if err != nil {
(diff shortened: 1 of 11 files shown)


m-rcl:
@mattklein123 @ysawa0 could one of you please have a look? Thanks in advance!

src/config_check_cmd/main.go
```diff
@@ -15,21 +15,23 @@ import (
...
 	defer func() {
 		err := recover()
 		if err != nil {
 			fmt.Printf("error loading rate limit configs: %s\n", err.(error).Error())
 			os.Exit(1)
 		}
 	}()
-	statsManager := stats.NewStatManager(gostats.NewStore(gostats.NewNullSink(), false), settings.NewSettings())
-	config.NewRateLimitConfigImpl(allConfigs, statsManager)
+	statsManager := stats.NewStatManager(gostats.NewStore(gostats.NewNullSink(), mergeDomainConfigs), settings.NewSettings())
```
mattklein123, line 26: I think this is a typo?

src/service/ratelimit.go
```diff
@@ -75,12 +75,15 @@ func (this *service) reloadConfig(statsManager stats.Manager) {
 		files = append(files, config.RateLimitConfigToLoad{key, snapshot.Get(key)})
 	}
 
-	newConfig := this.configLoader.Load(files, statsManager)
+	this.configLock.Lock()
+	defer this.configLock.Unlock()
```
mattklein123, line 79: For performance reasons I think you want to keep the locking where it is. I think you can just move the settings construction outside the lock? I'm not sure why it's in there. Seems like that should be static anyway?

mattklein123 (changes requested):
LGTM with small comments, thank you.

src/service/ratelimit.go
m-rcl: Agree. Having a second look at it, moving out the constructor should not have any impact.

src/config_check_cmd/main.go
mattklein123, line 26:
```suggestion
	statsManager := stats.NewStatManager(gostats.NewStore(gostats.NewNullSink(), false), settings.NewSettings())
```

Not sure why this line changed?

mattklein123 (approved):
Thanks!

PR #360, opened 2022-09-09
Add switch to force config reload if files are deleted

m-rcl:
Motivation: Allow to trigger a configuration reload if any file within the config directory is deleted
Usecase: Having multiple Kubernetes Configmaps, containing ratelimit configurations. On resource deletion the corresponding configuration should be removed from the active ratelimit configuration as well. 

I'm not sure whether the switch introduced in this PR (RUNTIME_WATCH_DELETION) is the most suitable option here. I could imagine following configuration options:

- Pass a comma separated string "Read,Write,Delete..." 
- Pass a numeric representation (similar to unix file permissions) 1=Read, 2=Write, 4=Delete, 8=Chmod - might be a bit too obscure but easy to parse
- Pass each operation as separate boolean variable - might be too verbose

@mattklein123 do you have an opinion on which way to go? Or should we stick with the 'minimal' setup of this PR?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,7 +142,7 @@ docker_multiarch_push: docker_multiarch_image
 
 .PHONY: integration_tests
 integration_tests:
-	docker-compose --project-dir $(PWD)  -f integration-test/docker-compose-integration-test.yml up --build  --exit-code-from tester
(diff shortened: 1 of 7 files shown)


mattklein123:
Hmm I'm think we should just do this in all cases? Is there any case in which someone wouldn't want this to happen? I'm thinking we should just fix it to watch deletion and then just see if anyone complains? WDYT?

PR #361, opened 2022-09-12
Reload configuration when deleting config files

m-rcl:
Supersedes:  #360

Problem: Having multiple Kubernetes Configmaps, containing ratelimit configurations,  the corresponding configuration should be removed from the active ratelimit configuration on resource deletion.

@mattklein123 As discussed in #360


mattklein123 (approved):
Thanks!

PR #364, opened 2022-09-19
Upgrade golang.org/x/net and alpine base image

dio:
To fix: 

- CVE-2022-37434 (zlib: a heap-based buffer over-read or buffer overflow in inflate.c...)
- CVE-2022-27664 (golang: net/http: handle server errors after sending GOAWAY)

Signed-off-by: Dhi Aurrahman <dio@rockybars.com>


mattklein123 (approved):
Thanks!

PR #374, opened 2022-11-15
Expose Config structures outside package

arkodg:
* Allow downstream projects to reuse the config structure and generate the rate limit service config programatically w/o creating their own custom structures.

Signed-off-by: Arko Dasgupta <arko@tetrate.io>

### DIFF
--- a/src/config/config_impl.go
+++ b/src/config/config_impl.go
@@ -17,7 +17,7 @@ type yamlReplaces struct {
 	Name string
 }
 
-type yamlRateLimit struct {
+type YamlRateLimit struct {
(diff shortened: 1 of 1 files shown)


mattklein123:
@arkodg I tried re-running and I'm not sure if the DNS failure is spurious or the test points somewhere that no longer exists. Can you take a look when you have time?

arkodg:
@mattklein123 raised https://github.com/envoyproxy/ratelimit/pull/375 to fix the test failure

mattklein123 (approved):
Thanks!

PR #375, opened 2022-11-15
Use _imaps._tcp.gmail.com in srv tests

arkodg:
* Seeing tests failures in CI with the current SRV entry `_xmpp-server._tcp.gmail.com.`
https://github.com/envoyproxy/ratelimit/actions/runs/3466737811/jobs/5802374477#step:3:257

* What is odd is I also tried `_xmpp-server._tcp.google.com.` which is used in the upstream `net.LookupSRV` tests
https://cs.opensource.google/go/go/+/master:src/net/lookup_test.go;l=47;drc=2b59307ac21135ab8db58e08fb98211fbedbb10d?q=LookupSRV&ss=go%2Fgo but that failed as well

Signed-off-by: Arko Dasgupta <arko@tetrate.io>

### DIFF
--- a/test/srv/srv_test.go
+++ b/test/srv/srv_test.go
@@ -49,12 +49,11 @@ func TestServerStringsFromSevWhenSrvIsWellFormedButNotLookupable(t *testing.T) {
 }
 
 func TestServerStrings(t *testing.T) {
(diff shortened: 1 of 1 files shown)


mattklein123 (approved):
Thanks!

PR #387, opened 2022-12-19
Print requested domain with descriptors in debug log

upgle:
If only the descriptor is printed in the log, it is not possible to know which domain the request is for.
Print domain in the log to make analysis easier.

```go
limitsToCheck[i] = snappedConfig.GetLimit(ctx, request.Domain, descriptor)
```


upgle:
re-open

PR #389, opened 2023-01-08
Include value in metrics for unspecified value

### LINKED ISSUE #181: [Proposal] detailed metric mode, for value in descriptor KEY_VALUE pair, if value not explicitly defined/matched in config

jespersoderlund:
Fixes #181 by adding a configuration key to a ratelimit descriptor to allow it to include the value in a statistics value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@
     - [Rate limit definition](#rate-limit-definition)
     - [Replaces](#replaces)
     - [ShadowMode](#shadowmode)
+    - [Including detailed metrics for unspecified values](#including-detailed-metrics-for-unspecified-values)
     - [Examples](#examples)
       - [Example 1](#example-1)
       - [Example 2](#example-2)
@@ -24,6 +25,7 @@
       - [Example 5](#example-5)
       - [Example 6](#example-6)
       - [Example 7](#example-7)
+      - [Example 8](#example-8)
   - [Loading Configuration](#loading-configuration)
   - [Log Format](#log-format)
   - [GRPC Keepalive](#grpc-keepalive)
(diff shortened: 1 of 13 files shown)


jespersoderlund:
Renamed the config-parameter to "detailed_metric" instead of the crazy-long config parameter I used first

README.md
```diff
@@ -490,6 +499,43 @@ descriptors:
           unit: second
 ```
 
+#### Example 8
+
+In this example there of where the extended metrics for a key where there will be detailed metrics even for the unknown values.
```
mattklein123, line 504: Typo? Can't parse.

mattklein123 (changes requested):
LGTM with small comment nit, thanks.

README.md
jespersoderlund: Yeah, not very clear, fixed

mattklein123 (approved):
Thanks!

marinasalmen:
Hey @mattklein123 I think something went wrong with the image build and push.
Could you maybe take a look at it? Looking forward for this feature! 🙌

mattklein123:
This should be fixed. There was a credential issue.

PR #398, opened 2023-02-18
Pass health-check when at least one config is loaded

PeterL328:
Pass health-check when at least one config is loaded.
We change the health to start in failed mode, and then once we load at least one config file go healthy. If we later fail to load any configs, go unhealthy again.


PeterL328:
@mattklein123 Can I get an approval to run the workflow?

PeterL328:
@ysawa0 I believe you are also a maintainer. Do you mind approving above workflow?

mattklein123 (changes requested):
At minimum this should be documented in the README, though I wonder if any of these should actually be config guarded? Thank you.

PeterL328:
> At minimum this should be documented in the README, though I wonder if any of these should actually be config guarded? Thank you.

Thanks for the review. Sure, I can make the new feature(at least one config loaded to be healthy) to be opt-ed in.

src/redis/driver_impl.go
```diff
@@ -36,7 +36,10 @@ func poolTrace(ps *poolStats, healthCheckActiveConnection bool, srv server.Serve
 				ps.connectionTotal.Add(1)
 				ps.connectionActive.Add(1)
 				if healthCheckActiveConnection && srv != nil {
-					srv.HealthCheckOK()
+					err := srv.HealthChecker().Ok(server.RedisHealthComponentName)
+					if err != nil {
+						logger.Errorf("Unable to update health status: %s", err)
+					}
```
mattklein123, line 42: This is new also, right? Seems like blocking on redis health should also be configurable as it's a change? Some for the other change that is not config guarded including sigterm? Sigterm seems OK to not config check, though it should be documented. But I think the redis change should be config guarded?

mattklein123 (changes requested):

slice-sarthakd (approved):

src/redis/driver_impl.go
PeterL328: With the "Config" health component config guarded, the behavior is the same as before. So I believe it should be fine to have the above not config guarded.
I can add more documentation about the redis and sigterm health components.

PeterL328: Let me know what you think :)

PR #403, opened 2023-03-16
Add wildcard support to descriptor values

### LINKED ISSUE #317: Use of wildcard in descriptor value
Hi,

We would like to use rate limit on header ":path"  for paths starting for example like: /api or /events

When we add a configuration like this:

```
- key: PATH
   value: /api/
   rate_limit:
       unit: minute
       requests_per_unit: 10

```
We notice that paths like "/api/feed" or "/api/whatever" are not included in the rule. Is there any way to set a wildcard  or similar to the descriptor value like: 

```
- key: PATH
   value: /api/*
   rate_limit:
       unit: minute
       requests_per_unit: 10

```
So all the paths under /api/.. go through the same rule?

Thanks

Pawka:
Fixes #317.

Add wildcard support to descriptor values to match endpoints under the value. This eliminates the need to explicitly list all endpoints to which rate limiter should be applied.

Example configuration:
```yaml
domain: test-domain
descriptors:
  - key: wild
    value: foo*
    rate_limit:
      unit: minute
      requests_per_unit: 20
```

### DIFF
--- a/src/config/config_impl.go
+++ b/src/config/config_impl.go
@@ -305,6 +305,15 @@ func (this *rateLimitConfigImpl) GetLimit(
 		finalKey := entry.Key + "_" + entry.Value
 		logger.Debugf("looking up key: %s", finalKey)
 		nextDescriptor := descriptorsMap[finalKey]
+
+		if nextDescriptor == nil {
+			for descriptorKey := range descriptorsMap {
+				if descriptorKey[len(descriptorKey)-1:] == "*" && strings.HasPrefix(finalKey, strings.TrimSuffix(descriptorKey, "*")) {
+					nextDescriptor = descriptorsMap[descriptorKey]
+				}
+			}
+		}
+
 		if nextDescriptor == nil {
 			finalKey = entry.Key
 			logger.Debugf("looking up key: %s", finalKey)
--- a/src/config/config_impl.go
+++ b/src/config/config_impl.go
@@ -40,8 +40,9 @@ type YamlRoot struct {
 }
 
 type rateLimitDescriptor struct {
-	descriptors map[string]*rateLimitDescriptor
-	limit       *RateLimit
+	descriptors  map[string]*rateLimitDescriptor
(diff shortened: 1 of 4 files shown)


Pawka:
Ha, it is interesting, since part of this PR was not created by myself but by @alexclarkofficial hence I can not sign in behalf of him.

@ysawa0 @mattklein123 what are your recommendations how to approach this case?

alexclarkofficial:
I can look into signing this later today.

Pawka:
@alexclarkofficial could you please either confirm my suggestion listed in the above commend is fine or sign off your change? Thanks.

alexclarkofficial:
Yep, you can sign for me.

Pawka:
> Yep, you can sign for me.

Thanks for confirmation!

Pawka:
@mattklein123 @ysawa0 Could you please approve the workflow and review the PR when you'll have time? Thanks!

src/config/config_impl.go
```diff
@@ -305,6 +305,15 @@ func (this *rateLimitConfigImpl) GetLimit(
...
 		logger.Debugf("looking up key: %s", finalKey)
 		nextDescriptor := descriptorsMap[finalKey]
+
+		if nextDescriptor == nil {
+			for descriptorKey := range descriptorsMap {
+				if descriptorKey[len(descriptorKey)-1:] == "*" && strings.HasPrefix(finalKey, strings.TrimSuffix(descriptorKey, "*")) {
+					nextDescriptor = descriptorsMap[descriptorKey]
+				}
+			}
+		}
```
mattklein123, line 315: Instead of doing this on every lookup, can you pre-store whether value has the potential for a wildcard and only do this in that case? That will avoid the perf hit for all users.

Pawka: @mattklein123 Updated the PR. Please review and let me know if this is what you've expected to see.

mattklein123 (approved):
LGTM, thanks!

PR #406, opened 2023-03-31
Add possibility to provide additional headers to grpc request in xds sotw provider

Automaat:
Adding headers to GRPC requests can be used in multiple ways, one and most important is adding Authorization data to requests between ratelimit service and config provider.

### DIFF
--- a/src/settings/settings.go
+++ b/src/settings/settings.go
@@ -68,6 +68,9 @@ type Settings struct {
 	// GrpcClientTlsSAN is the SAN to validate from the client cert during mTLS auth
 	ConfigGrpcXdsServerTlsSAN string `envconfig:"CONFIG_GRPC_XDS_SERVER_TLS_SAN" default:""`
 
+	// xDS grpc client configuration
+	ConfigGrpcXdsClientAdditionalHeaders map[string]string `envconfig:"CONFIG_GRPC_XDS_CLIENT_ADDITIONAL_HEADERS"`
+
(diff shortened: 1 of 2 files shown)


src/settings/settings.go
```diff
@@ -68,6 +68,9 @@ type Settings struct {
 	// GrpcClientTlsSAN is the SAN to validate from the client cert during mTLS auth
 	ConfigGrpcXdsServerTlsSAN string `envconfig:"CONFIG_GRPC_XDS_SERVER_TLS_SAN" default:""`
 
+	// xDS grpc client configuration
+	ConfigGrpcXdsClientAdditionalHeaders map[string]string `envconfig:"CONFIG_GRPC_XDS_CLIENT_ADDITIONAL_HEADERS"`
```
mattklein123, line 72: Please document this in the README with the other options. Also, please add a test if possible. Thank you.

mattklein123 (changes requested):

src/settings/settings.go
Automaat: Documented this in README. Unfortunately, I was unable to add any tests. I thought about testing it against echo server, but this would end up with an invalid `DiscoveryResponse` from server and I wouldn't have access to the response in test to check if proper headers were returned. If you have any ideas how to test this in the current setup, let me know.

mattklein123 (approved):
Thanks!

PR #420, opened 2023-05-09
fix wildcard support to cover nested descriptors

amiramw:
each descriptor level can have a wildcard property therefore GetLimit should inspect there and not only on the base level


amiramw:
@mattklein123 I’d really appreciate it if you could take a look

mattklein123 (approved):

Issue #423, opened 2023-05-18
Bug: Invalid shadow metric when local cache hit


upgle:
## Description

When sending requests at the same rate, the shadow mode metric does not give the exact number of requests that should be blocked (shadowed). Even the over limit metric does not work properly with shadow mode.

Looking for the cause, I've found that the metric count does not increase when the local cache is hit.
And I noticed that some of the test code was also wrong.

#### Invalid Metric
- ratelimit_service_rate_limit_over_limit
    - The number of blocked(including shadow_mode) requests
- ratelimit_service_rate_limit_shadow_mode
    - The number of requests that should have been blocked, but the response was OK (shadow)

<img width="1572" alt="image" src="https://github.com/envoyproxy/ratelimit/assets/5635513/0eec0852-7fc2-46f0-924d-8861594c5d76">

#### Expected

The shadow mode metric should show the actual number of shadowed(blocked but returns OK) requests.

<img width="1105" alt="image" src="https://github.com/envoyproxy/ratelimit/assets/5635513/79c317a9-43fc-4a4f-91f4-69d7d1ba5fc4">

PR #424, opened 2023-05-18
Fix shadow mode metric is not collected when local cache hit

upgle:
## Description

This PR fixes https://github.com/envoyproxy/ratelimit/issues/423 issue where statistics were not collected correctly when hitting the local cache.

If the local cache is hit, the rate limit service does not perform the INCR command, but the GetResponseDescriptorStatus method does not handle it properly.

### Invalid metric
shadow mode stats are not collected for the actual number of requests.

image

### Expected
shadow mode stat should be the same value as over limit .

image

### DIFF
--- a/src/limiter/base_limiter.go
+++ b/src/limiter/base_limiter.go
@@ -80,9 +80,9 @@ func (this *BaseRateLimiter) GetResponseDescriptorStatus(key string, limitInfo *
 			nil, 0)
 	}
 	var responseDescriptorStatus *pb.RateLimitResponse_DescriptorStatus
-	over_limit := false
+	isOverLimit := false
 	if isOverLimitWithLocalCache {
-		over_limit = true
+		isOverLimit = true
 		limitInfo.limit.Stats.OverLimit.Add(uint64(hitsAddend))
 		limitInfo.limit.Stats.OverLimitWithLocalCache.Add(uint64(hitsAddend))
 		responseDescriptorStatus = this.generateResponseDescriptorStatus(pb.RateLimitResponse_OVER_LIMIT,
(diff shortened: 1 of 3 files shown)


test/redis/fixed_cache_impl_test.go
```diff
@@ -579,15 +580,17 @@ func TestOverLimitWithLocalCacheShadowRule(t *testing.T) {
...
+	// Even if you hit the local cache, other metrics should increase normally.
 	assert.Equal(uint64(4), limits[0].Stats.TotalHits.Value())
-	assert.Equal(uint64(1), limits[0].Stats.OverLimit.Value())
-	assert.Equal(uint64(0), limits[0].Stats.OverLimitWithLocalCache.Value())
+	assert.Equal(uint64(2), limits[0].Stats.OverLimit.Value())
+	assert.Equal(uint64(1), limits[0].Stats.OverLimitWithLocalCache.Value())
+	assert.Equal(uint64(2), limits[0].Stats.ShadowMode.Value())
 	assert.Equal(uint64(1), limits[0].Stats.NearLimit.Value())
-	assert.Equal(uint64(3), limits[0].Stats.WithinLimit.Value())
+	assert.Equal(uint64(2), limits[0].Stats.WithinLimit.Value())
```
upgle, line 593: This value should be 2. Please refer to the original test code.
(Even if shadow mode is enabled, this value should not be changed.)

https://github.com/envoyproxy/ratelimit/blob/b3562caaa3393766cebeb69e34281d3565154c56/test/redis/fixed_cache_impl_test.go#L280

src/redis/fixed_cache_impl.go
```diff
@@ -61,14 +61,12 @@ func (this *fixedRateLimitCacheImpl) DoLimit(
...
 		if this.baseRateLimiter.IsOverLimitWithLocalCache(cacheKey.Key) {
-
 			if limits[i].ShadowMode {
 				logger.Debugf("Cache key %s would be rate limited but shadow mode is enabled on this rule", cacheKey.Key)
 			} else {
 				logger.Debugf("cache key is over the limit: %s", cacheKey.Key)
-				isOverLimitWithLocalCache[i] = true
 			}
-
+			isOverLimitWithLocalCache[i] = true
```
upgle, line 69: Shadow mode and local cache should be irrelevant.

src/limiter/base_limiter.go
```diff
@@ -80,9 +80,9 @@ func (this *BaseRateLimiter) GetResponseDescriptorStatus(key string, limitInfo *
 			nil, 0)
 	}
 	var responseDescriptorStatus *pb.RateLimitResponse_DescriptorStatus
-	over_limit := false
```
upgle, line 83: This does not follow the Golang naming convention.

mattklein123:
Can you check CI? Not sure if it's a flake or not.

upgle:
@mattklein123 
All tests passed in my environment. Could you please re-run the PR build again?
The error log shows that the redis connection was refused.

mattklein123 (approved):

Issue #427, opened 2023-06-02
Stop incrementing cache keys when overall status is over limited


chashikajw:
## Scenario

Let's say we have grouped routes

1. Group: Foo
   -  /foo/v1/hello-1
2. Group: Bar
   - /bar/v1/hello-1
   - /bar/v1/hello-special

And we need to apply the following rate limits for the group: Bar.

1. Per group -> **10** req/min
2. Special rate limit for the route **/bar/v1/hello-special** in group Bar -> **3** req/min

The expectation is,

1. if we invoke **/bar/v1/hello-special** route more than 3 times it should be rate limited.
2. Then the other route(**/bar/v1/hello-1**) has still 7 req/min quota. 
3. Although we invoke **/bar/v1/hello-special** route 5 times,  that should be rate limited from 3 and for the other 2 invocations it should return 429, and **/bar/v1/hello-1** route still should have 7 req/min quota)

But the current behavior is,

1. If we invoke **/bar/v1/hello-special** route 5 times it is rate limited from the 3 requests and for the next 2 invocations, it returns 429 responses. (Overall 5 invocations but only 3 are valid increments)
2. Then if we invoke **/bar/v1/hello-1** route, it should have the 7 req/min quota. But it has only a 5 req/min quota since we invoke **/bar/v1/hello-special** route 5 times including 2 invalid invocations which return 429 responses.

I believe this behavior should be fixed and need to stop incrementing cache keys when the overall status is over-limited.

## Reproducing steps

**Proxy yaml**

```
- match:
    prefix: /foo/v1/hello-1
  route:
    cluster: mock
    rate_limits:
      - actions:
          - generic_key:
              descriptor_key: "group"
              descriptor_value: "foo"


- match:
    prefix: /bar/v1/hello-1
  route:
    cluster: mock
    rate_limits:
      - actions:
          - generic_key:
              descriptor_key: "group"
              descriptor_value: "bar"
- match:
    prefix: /bar/v1/hello-special
  route:
    cluster: mock
    rate_limits:
      - actions:
          - generic_key:
              descriptor_key: "group"
              descriptor_value: "bar"
      - actions:
          - generic_key:
              descriptor_key: "special-policy"
              descriptor_value: "3PerMin"

```

**Rate Limit Service Config Yaml**

```
descriptors:
  - key: group
    rate_limit:
      unit: minute
      requests_per_unit: 10
  - key: special-policy
    value: 3PerMin
    rate_limit:
      unit: minute
[...]

PR #430, opened 2023-06-06
Fix incrementing cache keys when overall status is over limited

chashikajw:
## Description
This will add a fix to stop incrementing cache keys when the overall status is over-limited.

### Fixes
- https://github.com/envoyproxy/ratelimit/issues/427

### DIFF
--- a/src/redis/fixed_cache_impl.go
+++ b/src/redis/fixed_cache_impl.go
@@ -53,32 +53,27 @@ func (this *fixedRateLimitCacheImpl) DoLimit(
 	results := make([]uint32, len(request.Descriptors))
 	var pipeline, perSecondPipeline Pipeline
 
-	for _, cacheKey := range cacheKeys {
-		if cacheKey.Key == "" {
-			continue
-		}
-
-		// Check if any of the descriptors are over limit in local cache.
-		if this.baseRateLimiter.IsOverLimitWithLocalCache(cacheKey.Key) {
-			hitsAddend = 0
-			break
-		}
-	}
-
-	// Now, actually setup the pipeline, skipping empty cache keys.
-	for i, cacheKey := range cacheKeys {
+	for j, cacheKey := range cacheKeys {
 		if cacheKey.Key == "" {
 			continue
 		}
 
 		// Check if key is over the limit in local cache.
 		if this.baseRateLimiter.IsOverLimitWithLocalCache(cacheKey.Key) {
-			if limits[i].ShadowMode {
+			hitsAddend = 0
+			if limits[j].ShadowMode {
 				logger.Debugf("Cache key %s would be rate limited but shadow mode is enabled on this rule", cacheKey.Key)
 			} else {
 				logger.Debugf("cache key is over the limit: %s", cacheKey.Key)
 			}
-			isOverLimitWithLocalCache[i] = true
+			isOverLimitWithLocalCache[j] = true
+			continue
+		}
+	}
+
+	// Now, actually setup the pipeline, skipping empty cache keys.
+	for i, cacheKey := range cacheKeys {
+		if cacheKey.Key == "" {
 			continue
 		}
 
--- a/src/redis/fixed_cache_impl.go
+++ b/src/redis/fixed_cache_impl.go
@@ -53,6 +53,18 @@ func (this *fixedRateLimitCacheImpl) DoLimit(
 	results := make([]uint32, len(request.Descriptors))
 	var pipeline, perSecondPipeline Pipeline
 
+	for _, cacheKey := range cacheKeys {
+		if cacheKey.Key == "" {
+			continue
+		}
+
+		// Check if any of the descriptors are over limit in local cache.
+		if this.baseRateLimiter.IsOverLimitWithLocalCache(cacheKey.Key) {
+			hitsAddend = 0
+			break
+		}
+	}
+
 	// Now, actually setup the pipeline, skipping empty cache keys.
 	for i, cacheKey := range cacheKeys {
 		if cacheKey.Key == "" {
--- a/test/redis/fixed_cache_impl_test.go
+++ b/test/redis/fixed_cache_impl_test.go
@@ -274,8 +274,8 @@ func TestOverLimitWithLocalCache(t *testing.T) {
 		},
 		cache.DoLimit(context.Background(), request, limits))
 	assert.Equal(uint64(4), limits[0].Stats.TotalHits.Value())
-	assert.Equal(uint64(2), limits[0].Stats.OverLimit.Value())
-	assert.Equal(uint64(1), limits[0].Stats.OverLimitWithLocalCache.Value())
+	assert.Equal(uint64(1), limits[0].Stats.OverLimit.Value())
+	assert.Equal(uint64(0), limits[0].Stats.OverLimitWithLocalCache.Value())
 	assert.Equal(uint64(1), limits[0].Stats.NearLimit.Value())
 	assert.Equal(uint64(2), limits[0].Stats.WithinLimit.Value())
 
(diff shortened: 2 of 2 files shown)


chashikajw:
Hi @mattklein123 ,

Kind reminder on this.

Thank you

src/settings/settings.go
```diff
@@ -83,11 +83,12 @@ type Settings struct {
...
-	LocalCacheSizeInBytes      int     `envconfig:"LOCAL_CACHE_SIZE_IN_BYTES" default:"0"`
-	NearLimitRatio             float32 `envconfig:"NEAR_LIMIT_RATIO" default:"0.8"`
-	CacheKeyPrefix             string  `envconfig:"CACHE_KEY_PREFIX" default:""`
-	BackendType                string  `envconfig:"BACKEND_TYPE" default:"redis"`
+	ExpirationJitterMaxSeconds        int64   `envconfig:"EXPIRATION_JITTER_MAX_SECONDS" default:"300"`
+	LocalCacheSizeInBytes             int     `envconfig:"LOCAL_CACHE_SIZE_IN_BYTES" default:"0"`
+	NearLimitRatio                    float32 `envconfig:"NEAR_LIMIT_RATIO" default:"0.8"`
+	CacheKeyPrefix                    string  `envconfig:"CACHE_KEY_PREFIX" default:""`
+	BackendType                       string  `envconfig:"BACKEND_TYPE" default:"redis"`
+	IncrementingCacheKeysForOverLimit bool    `envconfig:"INCREMENTING_CACHE_KEYS_FOR_OVER_LIMIT" default:"true"`
```
mattklein123, line 91: I don't understand what this change does. Can you add documentation? Is this a bug that should just be fixed?

mattklein123 (changes requested):

src/settings/settings.go
chashikajw: Hi @mattklein123,

I have added a new setting called **IncrementingCacheKeysForOverLimit**. The reasons for this is to keep the backward compatibility and in order to fix the bug that I mentioned in the issue https://github.com/envoyproxy/ratelimit/issues/427

I will add the documentation also to get a clear understanding of the behavior if we use IncrementingCacheKeysForOverLimit

Thank you

chashikajw: This is what I am fixing from this Pr

Let's say we have grouped routes

1. Group: Foo
   -  /foo/v1/hello-1
2. Group: Bar
   - /bar/v1/hello-1
   - /bar/v1/hello-special

And we need to apply the following rate limits for the group: Bar.

1. Per group -> **10** req/min
2. Special rate limit for the route **/bar/v1/hello-special** in group Bar -> **3** req/min

The expectation is,

1. if we invoke **/bar/v1/hello-special** route more than 3 times it should be rate limited.
2. Then the other route(**/bar/v1/hello-1**) has still 7 req/min quota. 
3. Although we invoke **/bar/v1/hello-special** route 5 times,  that should be rate limited from 3 and for the other 2 invocations it should return 429, and **/bar/v1/hello-1** route still should have 7 req/min quota)

But the current behavior is,

1. If we invoke **/bar/v1/hello-special** route 5 times it is rate limited from the 3 requests and for the next 2 invocations, it returns 429 responses. (Overall 5 invocations but only 3 are valid increments)
2. Then if we invoke **/bar/v1/hello-1** route, it should have the 7 req/min quota. But it has only a 5 req/min quota since we invoke **/bar/v1/hello-special** route 5 times including 2 invalid invocations which return 429 responses.

mattklein123: OK this makes sense, though I wonder if this should just be the default. If you want to keep the option please document it in README.md with the other options.

chashikajw: Hi @mattklein123,

I document the **INCREMENTING_CACHE_KEYS_FOR_OVER_LIMIT** config from 96185143ba0888f71198ee8f406e572b07d0b86a

Thank you

README.md
```diff
@@ -885,8 +885,11 @@ You can specify the debug server address with the `DEBUG_HOST` and `DEBUG_PORT`
 # Local Cache
 
 Ratelimit optionally uses [freecache](https://github.com/coocood/freecache) as its local caching layer, which stores the over-the-limit cache keys, and thus avoids reading the
-redis cache again for the already over-the-limit keys. The local cache size can be configured via `LocalCacheSizeInBytes` in the [settings](https://github.com/envoyproxy/ratelimit/blob/master/src/settings/settings.go).
+redis cache again for the already over-the-limit keys. 
+Cache configuration options can be set using the following environment variables:
+1. The local cache size can be configured via `LocalCacheSizeInBytes` in the [settings](https://github.com/envoyproxy/ratelimit/blob/master/src/settings/settings.go).
 If `LocalCacheSizeInBytes` is 0, local cache is disabled.
+2. To control the behavior of cache keys when the overall status is over-limited, set the `INCREMENTING_CACHE_KEYS_FOR_OVER_LIMIT` variable to "false". This is particularly useful when utilizing multiple descriptors for a group of routes. Note that for this configuration to work, the local cache must be enabled.
```
mattklein123, line 892: Sorry from this comment as a user I don't know what this does or why I would set it. Can you expand? I'm still not convinced this isn't just a bug fix. Does it actually need a config flag? What would be the situation in which someone would want to keep this as true?

chashikajw: Hi @mattklein123,

We don't actually need the config. I added that config only to keep the current behavior as it is (assuming that someone else has depended on that behavior, although it is a bug). But I also believe we don't need this config and we can directly fix this in the code below. Shall I do it?

We can directly execute the code https://github.com/envoyproxy/ratelimit/pull/430/files#diff-cf4d55b6b42859dee0ecf1d127327f14a09781d93dc53cd6267019cb191c870cR61.

Thank you

chashikajw: Hi @mattklein123 ,

Just fixed the bug from c39e74da49776669ce0b74b5c5d0b9926a8f52e8

Thank you

mattklein123 (changes requested):
Please add a test that covers this case.

chashikajw:
> Please add a test that covers this case.

I changed the test cases according to the bug fix from aa5631017bd1cb902d47a9c406b7395fef637134

mattklein123:
Test look broken. I'm not really sure how you can just change the values of the tests when they didn't fail last time.

chashikajw:
> Test look broken. I'm not really sure how you can just change the values of the tests when they didn't fail last time.

The corrected test cases was failed last time due to my change and I changed it according to the fix. I will further verify this test failures locally.

BLasan:
@freedomljc Hi, would you mind providing an overall idea on the failing asserts? If the fix is correct, then we can simply change assert values right? Otherwise, if you're trying to check some sort of a different scenario, could you please explain the anticipations of those assertions and whether the failures are expected or not?

src/redis/fixed_cache_impl.go
```diff
@@ -53,6 +53,18 @@ func (this *fixedRateLimitCacheImpl) DoLimit(
...
 
+	for _, cacheKey := range cacheKeys {
+		if cacheKey.Key == "" {
+			continue
+		}
+
+		// Check if any of the descriptors are over limit in local cache.
+		if this.baseRateLimiter.IsOverLimitWithLocalCache(cacheKey.Key) {
+			hitsAddend = 0
+			break
```
renuka-fernando, line 64: Isn't it possible to return from this function? We do not want to hit Redis right?

Tharsanan1:
@mattklein123 could you please review the latest changes and approve them? I have changed the integration test to test the correct behavior in Redis local cache flow [here](https://github.com/envoyproxy/ratelimit/pull/430/files#diff-ca62391b26e30e01d1f8cee0eac7829ba8e9b5e22d80eab0e83881086fb4177bR597)

mattklein123 (approved):

freedomljc:
Just back from PTO, catching up right now. 
@chashikajw, i think the PR would change the behavior for https://github.com/envoyproxy/ratelimit#example-8. By design, the threshold is not shared across different descriptors. I left a comment in https://github.com/envoyproxy/ratelimit/issues/427 as well.

Issue #440, opened 2023-08-16
Rate Limiting quota adjustment and key identification prior to rate limiting when applying multiple descriptors


chashikajw:
This issue is related to issue [1]. We have fixed the main issues from the PR [2], but it is not fully resolved. In order to fully address the increment issue, we should have a way to identify a key is rate limited before it rate limits.

## Scenario

Let's say we have grouped routes

1. Group: Foo
   -  /foo/v1/hello-1
2. Group: Bar
   - /bar/v1/hello-1
   - /bar/v1/hello-special

And we need to apply the following rate limits for the group: Bar.

1. Per group -> **10** req/min
2. Special rate limit for the route **/bar/v1/hello-special** in group Bar -> **3** req/min

The expectation is,

1. if we invoke **/bar/v1/hello-special** route more than 3 times it should be rate limited.
2. Then the other route(**/bar/v1/hello-1**) has  7 req/min quota. 

But the current behaviour is,

1. If we invoke the /bar/v1/hello-special route 6 times, it gets rate-limited after 3 requests. For the next 3 invocations, it returns 429 responses. (In total, there are 6 invocations, but only 3 are considered valid increments.)
2. If we then invoke the /bar/v1/hello-1 route, it should have a 7 req/min quota, but it only has a 6 req/min quota.
3. Similarly, the remaining quota is consistently 1 less than the expected remaining quota.

The issue lies in identifying whether a key is over the limit [3] only after it has been rate-limited, leading to the cessation of key increment. Instead, we need a way to identify this before its rate limited and stopping the key increment in advance.

[1] https://github.com/envoyproxy/ratelimit/issues/427
[2] https://github.com/envoyproxy/ratelimit/pull/430
[3] https://github.com/envoyproxy/ratelimit/blob/main/src/redis/fixed_cache_impl.go#L65

## Reproducing steps

**Proxy yaml**

```
- match:
    prefix: /foo/v1/hello-1
  route:
    cluster: mock
    rate_limits:
      - actions:
          - generic_key:
              descriptor_key: "group"
              descriptor_value: "foo"


- match:
    prefix: /bar/v1/hello-1
  route:
    cluster: mock
    rate_limits:
      - actions:
          - generic_key:
              descriptor_key: "group"
              descriptor_value: "bar"
- match:
    prefix: /bar/v1/hello-special
  route:
    cluster: mock
    rate_limits:
      - actions:
          - generic_key:
              descriptor_key: "group"
[...]

PR #443, opened 2023-08-21
Fix ratelimit counter issue when using multiple descriptors

chashikajw:
### Description
This PR will address following two issues. 
1. When using multiple descriptors, this will stop incrementing cache keys when overall status is over limited.
2. Instead of using local cache, this uses redis cache to fetch cache keys to ensure the accuracy of the result in multi-host environments

## Fixes
- https://github.com/envoyproxy/ratelimit/issues/440
- https://github.com/envoyproxy/ratelimit/issues/427.  **(Fix this issue without engaging local cache)**

### DIFF
--- a/README.md
+++ b/README.md
@@ -904,6 +904,12 @@ As well Ratelimit supports TLS connections and authentication. These can be conf
 1. `REDIS_AUTH` & `REDIS_PERSECOND_AUTH`: set to `"username:password"` to enable username-password authentication to the redis host.
 1. `CACHE_KEY_PREFIX`: a string to prepend to all cache keys
 
+For controlling the behavior of cache key incrementation when they reach the overlimit, you can use the following configuration:
+
+1. `STOP_CACHE_KEY_INCREMENT_WHEN_OVERLIMIT`: Set this configuration to `true` to disallow key incrementation when keys reach the overlimit.
+
+`STOP_CACHE_KEY_INCREMENT_WHEN_OVERLIMIT` is useful when multiple descriptors are included in a single request. Setting this to `true` can prevent the incrementation of other descriptors' counters if any of the descriptors reach the overlimit.
+
 ## Redis type
 
 Ratelimit supports different types of redis deployments:
--- a/README.md
+++ b/README.md
@@ -904,11 +904,11 @@ As well Ratelimit supports TLS connections and authentication. These can be conf
 1. `REDIS_AUTH` & `REDIS_PERSECOND_AUTH`: set to `"username:password"` to enable username-password authentication to the redis host.
 1. `CACHE_KEY_PREFIX`: a string to prepend to all cache keys
 
-For controlling the behavior of cache key incrementation when they reach the overlimit, you can use the following configuration:
+For controlling the behavior of cache key incrementation when any of them is already over the limit, you can use the following configuration:
 
-1. `STOP_CACHE_KEY_INCREMENT_WHEN_OVERLIMIT`: Set this configuration to `true` to disallow key incrementation when keys reach the overlimit.
+1. `STOP_CACHE_KEY_INCREMENT_WHEN_OVERLIMIT`: Set this configuration to `true` to disallow key incrementation when one of the keys is already over the limit.
 
-`STOP_CACHE_KEY_INCREMENT_WHEN_OVERLIMIT` is useful when multiple descriptors are included in a single request. Setting this to `true` can prevent the incrementation of other descriptors' counters if any of the descriptors reach the overlimit.
+`STOP_CACHE_KEY_INCREMENT_WHEN_OVERLIMIT` is useful when multiple descriptors are included in a single request. Setting this to `true` can prevent the incrementation of other descriptors' counters if any of the descriptors is already over the limit.
 
 ## Redis type
 
--- a/src/redis/fixed_cache_impl.go
+++ b/src/redis/fixed_cache_impl.go
@@ -7,6 +7,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 
+	"github.com/envoyproxy/ratelimit/src/settings"
 	"github.com/envoyproxy/ratelimit/src/stats"
 
 	"github.com/coocood/freecache"
@@ -19,7 +20,10 @@ import (
 	"github.com/envoyproxy/ratelimit/src/utils"
 )
 
-var tracer = otel.Tracer("redis.fixedCacheImpl")
+var (
+	tracer     = otel.Tracer("redis.fixedCacheImpl")
+	rlSettings = settings.NewSettings()
+)
 
 type fixedRateLimitCacheImpl struct {
 	client Client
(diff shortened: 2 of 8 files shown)


chashikajw:
Hi @mattklein123,

Could you please review the PR.

Thank you

chashikajw:
Hi @mattklein123,

Could you please review the PR.

Thank you

chashikajw:
Hi @mattklein123,

Could you please review the PR.

Thank you

chashikajw:
Hi @mattklein123,

Kind reminder on this

Thank you

mattklein123:
@freedomljc do you have a sec to look at this since you commented on the other PR? I don't have time to do a real code review of this. Thank you.

src/redis/fixed_cache_impl.go
```diff
@@ -69,11 +79,33 @@ func (this *fixedRateLimitCacheImpl) DoLimit(
...
 			isOverLimitWithLocalCache[i] = true
-			hitAddendForRedis = 0
+			hitsAddendForRedis = 0
 			overlimitIndex = i
+			isCacheKeyOverlimit = true
 			continue
+		} else {
+			// Only if none of the cache key is over limit, call redis to check whether it is getting overlimited.
+			if !isCacheKeyOverlimit {
+				if this.perSecondClient != nil && cacheKey.PerSecond {
```
freedomljc, line 89: Looks like that it might send multiple requests to redis first for overlimit verification purpose, for every single ratelimit request. Maybe we should balance on the accuracy on this specific use case VS the performance impact. 

Alternatively, we can have a feature flag regarding whether to enable overlimit verification. Besides, when enabling overlimit verification, probably we can use batchGet or pipeline to reduce the perf impact.

chashikajw: Yes I will do the suggested approaches to improve the perf.  BTW in previous PR, we have discussed to not to have that feature flag. Anyway when considering the perf impact, I also think its better to have that flag although we considered this as a bug. WDYT @mattklein123 ? Shall we have the feature flag[1] ?

[1] https://github.com/envoyproxy/ratelimit/pull/430#discussion_r1231461436

mattklein123: If there are concerns I would just have a config knob and clearly document it. Thank you.

chashikajw: @freedomljc , @mattklein123 Added the optimization and perf improvement in 0184b1abc6e72f9d7ea1ee939f0733338c6bbc77 and added the new config and documentation in 42ffea181c2849a7014dd89a680ad07e4dda0ac1

chashikajw:
Hi @mattklein123, @freedomljc 

Kind reminder on this

Thank you

src/redis/fixed_cache_impl.go
```diff
@@ -51,29 +59,96 @@ func (this *fixedRateLimitCacheImpl) DoLimit(
...
+				}
+				// Now fetch the pipeline.
+				limitBeforeIncrease := currentCount[i]
+				limitAfterIncrease := limitBeforeIncrease + hitsAddend
+
+				limitInfo := limiter.NewRateLimitInfo(limits[i], limitBeforeIncrease, limitAfterIncrease, 0, 0)
+
+				if this.baseRateLimiter.IsOverLimitThresholdReached(limitInfo) {
+					hitsAddendForRedis = 0
+					nearlimitIndex = i
```
freedomljc, line 125: Seems it only handles the case when there's only one over-the-limit descriptor. If there are multiple over-the-limit descriptors at the same time, will it continue to increment the counter in other over-the-limit descriptors?

freedomljc (commented):
Can we also add tests for the case when `StopCacheKeyIncrementWhenOverlimit` is true?

src/redis/fixed_cache_impl.go
chashikajw: Here if it has multiple discriptors which has same thresholds, if first one in the list is identified as it in the nearlimit, it is stop incrementing other descriptors which including the other descriptor which has same threshold also.

That mean 1 is remaining for the other descriptor which has same threshold although one we identified as in the nearlimit is getting overlimit in next request. To keep the consistency, will add a list to store and handle that same threshold scenario for multiple descriptors.

chashikajw: Handled multiple key scenario in 72b9230f584b5e1357d6cb4bea1f9aac782d8a50

chashikajw:
> Can we also add tests for the case when `StopCacheKeyIncrementWhenOverlimit` is true?

Seems like we need to pass the StopCacheKeyIncrementWhenOverlimit config to the  **fixedRateLimitCacheImpl** struct and need to modify the paramaters accordingly(https://github.com/envoyproxy/ratelimit/blob/main/src/redis/fixed_cache_impl.go#L145) to apply the integration and unit test with StopCacheKeyIncrementWhenOverlimit = true scenario . Hence this will need considerable change. Is it fine to do that ?
 https://github.com/envoyproxy/ratelimit/blob/main/src/redis/fixed_cache_impl.go#L145

README.md
```diff
@@ -904,6 +904,12 @@ As well Ratelimit supports TLS connections and authentication. These can be conf
 1. `REDIS_AUTH` & `REDIS_PERSECOND_AUTH`: set to `"username:password"` to enable username-password authentication to the redis host.
 1. `CACHE_KEY_PREFIX`: a string to prepend to all cache keys
 
+For controlling the behavior of cache key incrementation when they reach the overlimit, you can use the following configuration:
```
freedomljc, line 907: `when they reach the overlimit` -> `when any of them is already over the limit`

freedomljc (commented):
Looks mostly good to me, just some comments on the test/docoment.
For the document, since the scenario you are trying to fix is specifically for the case when any key is already over the limit, might be good to explicitly call it out.

test/redis/fixed_cache_impl_test.go
```diff
@@ -626,3 +626,113 @@ func TestRedisTracer(t *testing.T) {
...
+			[]*pb.RateLimitResponse_DescriptorStatus{
+				{Code: pb.RateLimitResponse_OVER_LIMIT, CurrentLimit: limits[0].Limit, LimitRemaining: 0, DurationUntilReset: utils.CalculateReset(&limits[0].Limit.Unit, timeSource)},
+				{Code: pb.RateLimitResponse_OVER_LIMIT, CurrentLimit: limits[1].Limit, LimitRemaining: 0, DurationUntilReset: utils.CalculateReset(&limits[1].Limit.Unit, timeSource)},
+			},
+			cache.DoLimit(context.Background(), request, limits))
+		assert.Equal(uint64(1), limits[0].Stats.TotalHits.Value())
+		assert.Equal(uint64(1), limits[0].Stats.OverLimit.Value())
+		assert.Equal(uint64(0), limits[0].Stats.NearLimit.Value())
+		assert.Equal(uint64(0), limits[0].Stats.WithinLimit.Value())
+		assert.Equal(uint64(1), limits[0].Stats.TotalHits.Value())
```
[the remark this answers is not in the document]
chashikajw: This is same test case set written for the testRedis function. Corrected to 1

test/redis/fixed_cache_impl_test.go
```diff
@@ -626,3 +626,113 @@ func TestRedisTracer(t *testing.T) {
...
+
+func testRedisWithStopCacheKeyIncrementWhenOverlimitConfig(usePerSecondRedis bool) func(*testing.T) {
+	return func(t *testing.T) {
+		assert := assert.New(t)
+		controller := gomock.NewController(t)
+		defer controller.Finish()
+		statsStore := gostats.NewStore(gostats.NewNullSink(), false)
+		sm := stats.NewMockStatManager(statsStore)
+
+		client := mock_redis.NewMockClient(controller)
```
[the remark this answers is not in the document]
chashikajw: Added comments to the unit tests

chashikajw:
Hi,

Kind reminder on this.

Thank you

mattklein123 (approved):

chashikajw:
Hi @mattklein123,

It seems like the corresponding Docker image hasn't been pushed due to an issue. What can we do about that?

Thank you

PR #450, opened 2023-09-28
feat: add optional custom response header with next reset timestamp

xvzf:
This adds an optional `X-Ratelimit-Reset` header, which contains a
timestamp on which the rate-limit window rolls over. It follows [Github API](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28).
design, which we adopted internally.
Hope to get this upstream :) If there's anything missing, let me know!

Signed-off-by: Matthias Riegler <matthias.riegler@ankorstore.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@
   - [Health Checking for Redis Active Connection](#health-checking-for-redis-active-connection)
 - [Memcache](#memcache)
 - [Custom headers](#custom-headers)
+  - [Optional custom headers (mimicking Github API behavior)](#optional-custom-headers-mimicking-github-api-behavior)
 - [Tracing](#tracing)
 - [mTLS](#mtls)
 - [Contact](#contact)
--- a/README.md
+++ b/README.md
@@ -1003,6 +1003,15 @@ The following environment variables control the custom response feature:
 1. `LIMIT_REMAINING_HEADER` - The default value is "RateLimit-Remaining", setting the environment variable will specify an alternative header name
 1. `LIMIT_RESET_HEADER` - The default value is "RateLimit-Reset", setting the environment variable will specify an alternative header name
 
+## Optional custom headers (mimicking Github API behavior)
+
+Some services such as Github add another header containing the timestamp on which the rate-limit clears. This can be configured using the following options:
+
+1. `LIMIT_RESET_TIMESTAMP_HEADER_ENABLED` - Enables the additional timestamp response header
+1. `LIMIT_RESET_TIMESTAMP_HEADER` - THe default value is "X-RateLimit-Reset" and defines the time at which the current rate limit window resets in UTC epoch seconds
+
+# Tracing
+
 You may use the following commands to quickly setup a openTelemetry collector together with a Jaeger all-in-one binary for quickstart:
 
 ```bash
@@ -1011,8 +1020,6 @@ otelcol-contrib --config examples/otlp-collector/config.yaml
 
 docker run -d --name jaeger -p 16686:16686 -p 14250:14250 jaegertracing/all-in-one:1.33
 ```
-
-# Tracing
 
 Ratelimit service supports exporting spans in OLTP format. See [OpenTelemetry](https://opentelemetry.io/) for more information.
 
--- a/src/service/ratelimit.go
+++ b/src/service/ratelimit.go
@@ -6,6 +6,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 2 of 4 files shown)


ChuckCrawford:
Maybe a silly question but why is this necessary in addition to the `DurationUntilReset` [already present on each Descriptor](https://github.com/ChuckCrawford/ratelimit/blob/0904ca1cd03d52e2641cf88c0ea776d873d70c9e/src/limiter/base_limiter.go#L196-L204) and the `rateLimitResetHeader` [already in the set of optional response headers](https://github.com/ChuckCrawford/ratelimit/blob/fb/brushing-off-lint/src/service/ratelimit.go#L225-L231) ?

xvzf:
Hey @ChuckCrawford 

I don't think your question is silly and I challenged the same internally.
The primary difference is the current `Ratelimit-Reset` header offers the _time for the rate limit to reset in seconds_.
The header introduced as part of this PR (`X-Ratelimit-Reset` is the **timestamp** _for the rate limit to reset in seconds_.

The information contained is the same hence I've added it behind a feature flag.

One of our teams adapted the rate-limit headers from Github, which includes both the `Retry-After` header (which maps to the `Ratelimit-Reset` header of this implementation) and the `X-Ratelimit-Reset` header (introduced in this PR).

Since it's a public API, we cannot drop a header as a client might've built logic around it.


Does this explain it?

xvzf:
Fixed the doc formating -- can you re-approve the run please @ChuckCrawford?

ChuckCrawford:
Hey @xvzf : ahhhh that makes sense. I am not an official maintainer or anything but I would be happy to give this PR a closer look early next week.

xvzf:
@ysawa0 can you check this when you have time by chance? :)

README.md
```diff
@@ -1003,6 +1004,15 @@ The following environment variables control the custom response feature:
 1. `LIMIT_REMAINING_HEADER` - The default value is "RateLimit-Remaining", setting the environment variable will specify an alternative header name
 1. `LIMIT_RESET_HEADER` - The default value is "RateLimit-Reset", setting the environment variable will specify an alternative header name
 
+## Optional custom headers (mimicking Github API behavior)
```
ChezCrawford, line 1007: If we move forward with this; the way the configuration is currently implemented feels a little awkward. In particular you always end up with both versions of the "reset header": 

```
RateLimit-Reset: 58
X-RateLimit-Reset: 1697219026
```

Would it make more sense to have a configuration option like `LIMIT_RESET_HEADER_TIMESTAMP_MODE` that defaults to `false` but when enabled; it writes the timestamp value to the header?

There would still be the potential awkwardness of the `RateLimit-Reset` header having a value that is not in accordance with the [draft-ietf-httpapi-ratelimit-headers-07](https://datatracker.ietf.org/doc/draft-ietf-httpapi-ratelimit-headers/) spec but the ability to rename the headers already exists. Maybe a comment in the docs and a warning-level log alerting users to this fact might suffice?

README.md
```diff
@@ -1003,6 +1004,15 @@ The following environment variables control the custom response feature:
...
+
+Some services such as Github add another header containing the timestamp on which the rate-limit clears. This can be configured using the following options:
+
+1. `LIMIT_RESET_TIMESTAMP_HEADER_ENABLED` - Enables the additional timestamp response header
+1. `LIMIT_RESET_TIMESTAMP_HEADER` - THe default value is "X-RateLimit-Reset" and defines the time at which the current rate limit window resets in UTC epoch seconds
+
+# Tracing
```
ChezCrawford, line 1014: nit: That tiny itch in my brain is so happy you fixed this.

src/service/ratelimit.go
```diff
@@ -258,18 +270,27 @@ func (this *service) rateLimitRemainingHeader(descriptor *pb.RateLimitResponse_D
 }
 
 func (this *service) rateLimitResetHeader(
-	descriptor *pb.RateLimitResponse_DescriptorStatus) *core.HeaderValue {
```
ChezCrawford, line 261: nit: I personally like sticking with the existing function style already used in this file.

ChezCrawford (commented):
I am not an official maintainer or anything but I gave this a look. It does seem to work but the configuration and resulting response feel a little awkward.

See https://github.com/envoyproxy/ratelimit/pull/450/files?diff=unified&w=0#r1358632182 for more detail.

xvzf:
thx for the comments - will get back to this probably end of the week @ChezCrawford

harpunius:
> thx for the comments - will get back to this probably end of the week @ChezCrawford

This would be cool to have. Any updated ETA?

xvzf:
Hey

Sorry for the delays, currently a bit busy with other projects -- will probably mid till end of December :/

harpunius:
Sorry for the bump, but this would be nice to have.

PR #452, opened 2023-10-03
Brushes off some lint

ChuckCrawford:
## Overview
* Resolves most go staticcheck errors
* Replaces a few deprecated packages / functions.
* Makes some minor changes to error responses on the `/json` API endpoint (see the tests in `test/server/server_impl_test.go` for more detail).

## Commentary
* Left [ST1006](https://staticcheck.io/docs/checks#ST1006) alone for now since `this` is used all over the place.
* Left the [server.runtime](https://github.com/envoyproxy/ratelimit/blob/a34dbf9eaf805817e3fc186b452d4717f74ef13c/src/server/server_impl.go#L63C2-L63C9) unused import alone since I was not 100% sure whether it was safe to remove.
* The minor changes to the error responses on the `/json` API endpoint were motivated by:
  * The deprecation of [github.com/golang/protobuf/jsonpb](https://pkg.go.dev/github.com/golang/protobuf/jsonpb) which is used to convert protobuf to JSON. Some of the error messages were coming from the internals of this library and were no longer available once replaced with [google.golang.org/protobuf/encoding/protojson](https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson).
  * It is generally a good security practice not to leak library internals back out to clients. [Though I am sure it is rare to expose something like this externally.]

### DIFF
--- a/src/server/server_impl.go
+++ b/src/server/server_impl.go
@@ -1,7 +1,6 @@
 package server
 
 import (
-	"bytes"
 	"context"
 	"expvar"
 	"fmt"
@@ -18,13 +17,13 @@ import (
 
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/keepalive"
+	"google.golang.org/protobuf/encoding/protojson"
 
 	"github.com/envoyproxy/ratelimit/src/provider"
 	"github.com/envoyproxy/ratelimit/src/stats"
 
 	"github.com/coocood/freecache"
 	pb "github.com/envoyproxy/go-control-plane/envoy/service/ratelimit/v3"
-	"github.com/golang/protobuf/jsonpb"
 	"github.com/gorilla/mux"
 	reuseport "github.com/kavu/go_reuseport"
 	"github.com/lyft/goruntime/loader"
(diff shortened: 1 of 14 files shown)


ysawa0:
Very nice. Just to confirm so these lints were all detected via tooling and safe to update, correct?

ChuckCrawford:
> Very nice. Just to confirm so these lints were all detected via tooling and safe to update, correct?

That is correct. Everything I changed was detected by [go staticcheck](https://staticcheck.io/). The only functional change here is very minor changes to the error responses of the `/json` API since the previous error responses were leaking library-specific details [and that library was deprecated].

src/server/server_impl.go
```diff
@@ -79,24 +78,28 @@ func (server *server) AddDebugHttpEndpoint(path string, help string, handler htt
...
+		body, err := io.ReadAll(request.Body)
+		if err != nil {
+			logger.Warnf("error: %s", err.Error())
+			writeHttpStatus(writer, http.StatusBadRequest)
+			return
+		}
+
+		if err := protojson.Unmarshal(body, &req); err != nil {
 			logger.Warnf("error: %s", err.Error())
-			http.Error(writer, err.Error(), http.StatusBadRequest)
```
ysawa0, line 92: Any particular reason we're removing the more descriptive error msgs in the responses? Is it for consistency, security?

ChuckCrawford: Ah my apologies! I have updated the PR description to provide more clarity around why I did this.

In this exact line you commented on we are still doing the same thing; I just refactored it into a shared function for consistency.

ysawa0 (approved):

PR #458, opened 2023-11-15
Add lefthand-side wildcard support to descriptor values

lukipro:
Current wildcard implementation doesn't allow to have a following configuration:
```
domain: example-domain
descriptors:
  - key: HOST
    value: "*.example.com"
    rate_limit:
      unit: minute
      requests_per_unit: 20
```

This is useful if one needs to rate limit based on dynamic subdomains, like:
`foo.example.com`, `bar.example.com`, `<unknown_yet>.example.com`

### DIFF
--- a/src/config/config_impl.go
+++ b/src/config/config_impl.go
@@ -315,13 +315,13 @@ func (this *rateLimitConfigImpl) GetLimit(
 
 		if nextDescriptor == nil && len(prevDescriptor.wildcardValues) > 0 {
 			for _, wildcardValue := range prevDescriptor.wildcardValues {
-				finalKey = entry.Key + "_" + wildcardValue
+				wildcardKey := entry.Key + "_" + wildcardValue
 				if strings.HasSuffix(entry.Value, strings.TrimPrefix(wildcardValue, "*")) {
-					nextDescriptor = descriptorsMap[finalKey]
+					nextDescriptor = descriptorsMap[wildcardKey]
 					break
 				}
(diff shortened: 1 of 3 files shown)


lukipro:
Hello, any chance to get this merged?

src/config/config_impl.go
```diff
@@ -313,9 +313,14 @@ func (this *rateLimitConfigImpl) GetLimit(
...
-			for _, wildcardKey := range prevDescriptor.wildcardKeys {
-				if strings.HasPrefix(finalKey, strings.TrimSuffix(wildcardKey, "*")) {
+		if nextDescriptor == nil && len(prevDescriptor.wildcardValues) > 0 {
+			for _, wildcardValue := range prevDescriptor.wildcardValues {
+				wildcardKey := entry.Key + "_" + wildcardValue
+				if strings.HasSuffix(entry.Value, strings.TrimPrefix(wildcardValue, "*")) {
+					nextDescriptor = descriptorsMap[wildcardKey]
+					break
+				}
+				if strings.HasPrefix(entry.Value, strings.TrimSuffix(wildcardValue, "*")) {
```
ysawa0, line 323: Are there any performance implications here due to the string comparison?

lukipro: I didn't check that for two reasons:
- I don't know how to perform a meaningful benchmark with regards to this piece of code
- Current code is already doing a string comparison in the same way

PR #466, opened 2024-01-05
setup ossf scorecard and codql workflows

mmorel-35:
Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,6 +1,7 @@
 name: Scorecard supply-chain security
 
-permissions: read-all
+permissions:
+  contents: read
 
 on:
   # For Branch-Protection check. Only the default branch is supported. See
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,68 @@
+name: Scorecard supply-chain security
+
+permissions: read-all
+
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
+  branch_protection_rule:
+  # To guarantee Maintained check is occasionally updated. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
+  schedule:
+    - cron: "31 17 * * 3"
+  push:
+    branches: ["main"]
+
+jobs:
+  analysis:
+    name: Scorecard analysis
+    runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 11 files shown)


mmorel-35:
cc @phlax ,
Noticed that there is a formatter for the yaml files.

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,68 @@
+name: Scorecard supply-chain security
+
+permissions: read-all
```
phlax, line 3: prefer just `contents`

phlax (commented):
@mmorel-35 thanks - same style nits - indents/ordering etc - but otherwise lgtm

phlax:
>  Noticed that there is a formatter for the yaml files.

yeah yamllint - we wrap it with `envoy.code.check` - would be good to add that and `glint` to all of the repos (maintainers willing) as we did in toolshed 

https://github.com/envoyproxy/toolshed/blob/fa87bda6b7edd8e426fded6587b5e773e45a026e/.github/workflows/ci.yml#L54-L63

mmorel-35:
@phlax would it be a lack of rights that explains the failure ?

phlax:
seems there are 2 issues - the `build` issue looks a lot to me like a network transient

not sure about `pre-commit` issue but it seems to have gone away in last push

phlax:
i dont think it can be permissions as such - its a `pull_request` trigger which is ~permissionless

mmorel-35:
OK ! Everything seems to work now.
Anything to add to merge this ?

phlax:
could you fix the indents please - othewise lgtm

cc @ysawa0 @mattklein123 would be good to get signoff from a repo maintainer

mmorel-35:
"Fixing" the indent means changing the way it is scope. It seems out of the scope of this PR. 
Can it be done in another PR ?

integration-test/Dockerfile.tester
```diff
@@ -1,4 +1,4 @@
-FROM alpine:latest
+FROM alpine@sha256:51b67269f354137895d43f3b3d810bfacd3945438e94dc5ac55fdac340352f48
```
ysawa0, line 1: So moving forward we should always ref a sha for images and such?

ysawa0 (approved):
This looks great, thank you.

integration-test/Dockerfile.tester
phlax: yeah its generally preferable - but probs we can let dependabot or similar take care of it

PR #481, opened 2024-01-12
stats: sanitize metric names

wybczu:
If the configuration contains a key value with a `:` or `|`, e.g.

```
- key: path
  value: "/foo:*"
  rate_limit:
    unit: minute
    requests_per_unit: 20
```

the reported statsd metrics are malformed.

### DIFF
--- a/src/stats/manager_impl.go
+++ b/src/stats/manager_impl.go
@@ -5,6 +5,7 @@ import (
 	logger "github.com/sirupsen/logrus"
 
 	"github.com/envoyproxy/ratelimit/src/settings"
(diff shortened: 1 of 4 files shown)


ysawa0 (commented):
This makes sense. Could we add some tests for this?

wybczu:
> This makes sense. Could we add some tests for this?

Sure thing! I'll update the PR over the weekend.

wybczu:
@ysawa0 I updated the PR.

ysawa0 (approved):

Issue #502, opened 2024-02-09
xDS client doesn't handle stream closure with RST_STREAM


akondapuram:
When the xDS-server closes the stream and signals the xDS-client with `RST_STREAM` frame and error code `NO_ERROR`, [xDS client](https://github.com/envoyproxy/ratelimit/blob/9901a9bf860a796df97a864a50a8e4c7ac65a9de/src/provider/xds_grpc_sotw_provider.go#L67) present in the ratelimit service, logs the error message thousands of times per second. 

Error message that is logged indefinitely:
`level=error msg="Failed to receive configuration from xDS Management Server: rpc error: code = Internal desc = stream terminated by RST_STREAM with error code: NO_ERROR"`

Please see the screenshot below: 
![Screenshot 2024-02-09 at 11 28 46 AM](https://github.com/envoyproxy/ratelimit/assets/152326265/f44bd7b7-850f-45f0-b984-a90ca40414ba)

PR #503, opened 2024-02-09
Retry the grpc connection when there's an error

akondapuram:
This PR aims to fix the hot-looping problem described in [Issue#502](https://github.com/envoyproxy/ratelimit/issues/502)

The issue here was when the xDS-server closed the stream, the xDS-client tried to NACK the previous response and it went berserk in a hot-loop trying to fetch the configuration updates from the closed stream. This happens because the [sotw.isConnErr](https://github.com/envoyproxy/ratelimit/blob/1a686864369fa7b15d2c638ea641ad321bba9024/src/provider/xds_grpc_sotw_provider.go#L100) doesn't return true in this case when the server signals the client with the following error message 
`rpc error: code = Internal desc = stream terminated by RST_STREAM with error code: NO_ERROR`. 

Here's the [sotw.isConnErr()](https://github.com/envoyproxy/go-control-plane/blob/3405e27a3aa7ab375100f8c046f1c9e856199638/pkg/client/sotw/v3/client.go#L140) for reference. 

So the idea of this PR is to retry the connection whenever there is an error trying to fetch the config, instead of just expecting and handling just a few error codes.


akondapuram:
@mattklein123 @renuka-fernando Requesting your review on this PR. Thank you. 
cc: @ramaraochavali

src/provider/xds_grpc_sotw_provider.go
```diff
@@ -76,10 +78,25 @@ func (p *XdsGrpcSotwProvider) initXdsClient() {
...
 }
 
+func (p *XdsGrpcSotwProvider) getJitteredExponentialBackOffDuration() time.Duration {
+	b := &backoff.Backoff{
+		Min:    10 * time.Second,
+		Max:    10 * time.Minute,
+		Factor: 0.5,
+		Jitter: true,
+	}
+	logger.Infof("Retry attempt# %f", b.Attempt())
```
mikegrass, line 96: You'll want to move `b` outside of this function, otherwise you are resetting the backoff each time through the loop and you would only ever see one attempt.

akondapuram: Ah thanks for pointing this Mike. You're right.

src/provider/xds_grpc_sotw_provider.go
```diff
@@ -99,11 +115,8 @@ func (p *XdsGrpcSotwProvider) watchConfigs() {
 		resp, err := p.adsClient.Fetch()
 		if err != nil {
 			logger.Errorf("Failed to receive configuration from xDS Management Server: %s", err.Error())
-			if sotw.IsConnError(err) {
-				p.retryGrpcConn()
-				return
-			}
-			p.adsClient.Nack(err.Error())
+			p.retryGrpcConn()
+			return
```
renuka-fernando, line 119: Shouldn't we retry gRPC conn only for connection errors?

ramaraochavali: When we run Xds Server behind Envoy, during pod shutdowns/server enforced max connection age, the client gets RESET frame like "rpc error: code = Internal desc = stream terminated by RST_STREAM with error code: NO_ERROR" which is not treated as connection failure (It is RESET). Envoy -> XDS control plane does not just retry connection failures but retries with a backup on any error for the same reason https://github.com/envoyproxy/envoy/blob/49425f55aa9212a64b3390909160c41dc22ff349/source/extensions/config_subscription/grpc/grpc_stream.h#L50

This PR just mimicks the Envoy behaviour

renuka-fernando (approved):

akondapuram:
Thank you for the approval @renuka-fernando 
@mattklein123 Could you please take a look when you get some time? Thank you!

ramaraochavali:
@mattklein123  can you please take a look and merge this if this looks good to you?

akondapuram:
Thank you for your review @mattklein123. Updated per comments. Please take a second look when you're free. Thanks!

README.md
```diff
@@ -647,7 +647,12 @@ To enable this behavior set `MERGE_DOMAIN_CONFIG` to `true`.
 xDS Management Server is a gRPC server which implements the [Aggregated Discovery Service (ADS)](https://github.com/envoyproxy/data-plane-api/blob/97b6dae39046f7da1331a4dc57830d20e842fc26/envoy/service/discovery/v3/ads.proto).
 The xDS Management server serves [Discovery Response](https://github.com/envoyproxy/data-plane-api/blob/97b6dae39046f7da1331a4dc57830d20e842fc26/envoy/service/discovery/v3/discovery.proto#L69) with [Ratelimit Configuration Resources](api/ratelimit/config/ratelimit/v3/rls_conf.proto)
 and with Type URL `"type.googleapis.com/ratelimit.config.ratelimit.v3.RateLimitConfig"`.
+
 The xDS client in the Rate limit service configure Rate limit service with the provided configuration.
+In case of connection failures, the xDS Client retries the connection to the xDS server with exponential backoff and the backoff parameters are configurable.
+
+`XDS_CLIENT_BACKOFF_JITTER`: set to `"true"` to add jitter to the exponential backoff.
```
mattklein123, line 654: Please document all the configuration settings you added.

mattklein123 (changes requested):

mattklein123 (approved):

PR #520, opened 2024-02-25
fix: init gostats to not flood log

guilhem:
Default gostats store flood messages on stderr except if statd or if GOSTATS_LOGGING_SINK_DISABLED is set to "true". This commit init store in a unambiguous way.


guilhem:
ping :man_shrugging:

mattklein123 (approved):

PR #585, opened 2024-05-13
Initial integration of DogStatsD

JDeuce:
`godogstats` package is implemented to bridge `lyft/gostats` with [dogstatsd](https://docs.datadoghq.com/developers/dogstatsd/).

There is a new statsd sink implemented which will output the stats using dogstatsd protocol over UDP which allows tags.

Thus, a new `mogrification` feature is also added which lets us mogrify the stat from internal metric names, into a different name + tags, similar to what is done in the [prom-stats-exporter example](https://github.com/envoyproxy/ratelimit/blob/ca55e1b31d023d8b8aedb8a21ca3b769fb632e95/examples/prom-statsd-exporter/conf.yaml#L48-L54)

Note: there is some runtime cost associated with mogrifying the metrics but it should be minimal due to the existing in memory flushing architecture.


ChuckCrawford:
I apologize for making a design comment on an already implemented PR but what if instead of continuing to build support for specific company / vendor formats we instead implemented vendor-neutral [OTEL Metrics](https://opentelemetry.io/docs/specs/otel/metrics/) ?

src/service_cmd/runner/runner.go
```diff
@@ -43,6 +44,17 @@ func NewRunner(s settings.Settings) Runner {
...
+		sink, err := godogstats.NewSink(
+			godogstats.WithStatsdHost(s.StatsdHost),
+			godogstats.WithStatsdPort(s.StatsdPort),
+			godogstats.WithMogrifierFromEnv(s.UseDogStatsdMogrifiers))
+		if err != nil {
+			logger.Fatalf("Failed to create statsd sink: %v", err)
+		}
+		logger.Info("Stats initialized for dogstatsd")
+		store = gostats.NewStore(sink, false)
 	} else if s.UseStatsd {
```
mattklein123, line 58: Can you make it an error if someone sets both dogstatsd and statsd to true?

mattklein123 (dismissed):
Small comment otherwise LGTM.

ysawa0 (approved):

Issue #538, opened 2024-03-18
How to turn off gostats log


ShyunnY:
I use ratelimit in kubernetes, and when I checked the logs, I found a lot of log output about gostats.

**pod.yaml:**
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: "ratelimit-pod"
  namespace: default
  labels:
    app: "ratelimit"
spec:
  containers:
    - name: ratelimit-container
      image: envoyproxy/ratelimit:master
      command:
        - /bin/ratelimit
      env:
        - name: LOG_FORMAT
          value: json
        - name: LOG_LEVEL
          value: error
        - name: REDIS_SOCKET_TYPE
          value: tcp
        - name: REDIS_URL
          value: redis-service.default.svc.cluster.local:6379
        - name: RUNTIME_ROOT
          value: /data
        - name: RUNTIME_SUBDIRECTORY
          value: ratelimit
        - name: RUNTIME_WATCH_ROOT
          value: "false"
        - name: USE_STATSD
          value: "false"
        - name: MERGE_DOMAIN_CONFIG
          value: "true"
        - name: TRACING_ENABLED
          value: "true"
        - name: TRACING_EXPORTER_PROTOCOL
          value: "http"
        - name: TRACING_SERVICE_NAMESPACE
          value: "production"
        - name: TRACING_SERVICE_INSTANCE_ID
          value: "ratelimit-envoy"
        - name: OTEL_EXPORTER_OTLP_ENDPOINT
          value: "xxx"
      ports:
        - containerPort: 8080
        - containerPort: 8081
        - containerPort: 6070
      volumeMounts:
        - name: ratelimit-config
          mountPath: /data/ratelimit/config
          subPath: config
  volumes:
    - name: ratelimit-config
      configMap:
        name: ratelimit-config
        items:
          - key: demo.yaml
            path: config/demo.yaml
          - key: demo1.yaml
            path: config/demo1.yaml
```

**k logs ratelimit-pod:**
```shell
$ k logs ratelimit-pod
time="2024-03-18T05:38:05Z" level=info msg="TracerProvider initialized with following parameters: protocol: http, serviceName: RateLimit, serviceNamespace: production, serviceInstanceId: ratelimit-envoy, samplingRate: 1.000000"
{"level":"info","ts":1710740295.077808,"logger":"gostats.loggingsink","msg":"flushing counter","json":{"name":"runtime.load_attempts","type":"counter","value":"1.000000"}}
[...]

Issue #321, opened 2022-01-27
Ratelimit container is flooded with gostats logs


ashok243:
We are trying to deploy ratelimit container in GKE, I used the sample provided [here](https://github.com/istio/istio/blob/release-1.12/samples/ratelimit/rate-limit-service.yaml). But we are seeing the ratelimit container logs are flooded with gostats debug messages indicating `Flush() called, all stats would be flushed` followed by the messages that shows log messages for `flushing counter ratelimit...`, `flushing gauge ratelimit...` etc. 

Can someone please suggest what could be causing this issue?

PR #586, opened 2024-05-15
Bridge gostats logs to the ratelimit logging system

nacx:
Related to https://github.com/envoyproxy/ratelimit/issues/321 and https://github.com/envoyproxy/ratelimit/issues/538

This PR bridges the `gostats` logging sink to the ratelimit logger, so that the statsd logs can be enabled/disabled according to the `LOG_LEVEL` env var. The addition is mostly a copy of the private gostats logging sync, using the ratelimit logger to print the log statements.

### DIFF
--- a/src/stats/logger.go
+++ b/src/stats/logger.go
@@ -1,71 +1,28 @@
 package stats
 
 import (
-	"encoding/json"
-	"fmt"
 	"strconv"
 	"time"
 
-	gostats "github.com/lyft/gostats"
 	logger "github.com/sirupsen/logrus"
 )
 
-// loggingSink is a gostats.FlushableSink that logs all stats using
+const logTemplate = `{"level":"debug","ts":%[1]v,"logger":"gostats.loggingsink","msg":"flushing %[2]v","json":{"name":"%[3]v","type":"%[2]v","value":"%[4]v"}}`
+
+// LoggingSink is a gostats.FlushableSink that logs all stats using
 // the rate limit logger.
-type loggingSink struct {
-	now func() time.Time
-}
+type LoggingSink struct{}
 
-// NewLoggingSink logs the gostats metrics using the rate limit logger.
-// See the gostats.NewLoggingSink function for more details.
-func NewLoggingSink() gostats.FlushableSink {
-	return &loggingSink{now: time.Now}
-}
-
-type sixDecimalPlacesFloat float64
-
-func (f sixDecimalPlacesFloat) MarshalJSON() ([]byte, error) {
-	var ret []byte
-	ret = strconv.AppendFloat(ret, float64(f), 'f', 6, 64)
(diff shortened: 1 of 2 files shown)


vikaspalaskar (approved):

src/stats/logger.go
```diff
@@ -0,0 +1,71 @@
+package stats
```
mattklein123, line 1: It seems a shame to copy this entire file. I don't write go anymore, and I doubt gostats is maintained anymore, but is there any way to avoid doing this? Can we just use a much simpler format for this and write directly to our logging sync? I don't think the format needs to stay the same.

nacx: I agree if the library is not properly maintained we should probably stop using it, but that would be a much more involved change, as we should then consider not only the stdout output.

For the purpose of this PR of removing the noise on the stdout, I've updated the implementation to be very minimal, and in a way that allows us to change the format more easily to anything we consider more appropriate, but keeping the current JSON for now. WDYT?

src/stats/logger.go
```diff
@@ -0,0 +1,28 @@
...
+		logger.Debugf(logTemplate, strconv.FormatFloat(sec, 'f', 6, 64), typ, name, value)
+	}
+}
+
+// Implementation of the gostats.FlushableSink interface
+
+func (s *LoggingSink) FlushCounter(name string, value uint64) { s.log(name, "counter", float64(value)) }
+func (s *LoggingSink) FlushGauge(name string, value uint64)   { s.log(name, "gauge", float64(value)) }
+func (s *LoggingSink) FlushTimer(name string, value float64)  { s.log(name, "timer", value) }
+func (s *LoggingSink) Flush()                                 { s.log("", "all stats", 0) }
```
mattklein123, line 28: Sorry what does this part do? Are we going to end up logging counters and stuff?

nacx: This is actually the current behavior when users start the ratelimit service with `DisableStats = false && UseStatsd = false`. It initializes a default gostats logging sink that prints all those logs to the standard output.

There are several minor issues asking for a way to silence those logs when that config combo is used, and this PR attempts to provide that by bridging the gostats to the ratelimit logger instead of writing them directly to the stdout, to allow users leverage the LOG_LEVEL setting accordingly.

One could argue that if users want to silence those logs, they should just set `DisableStats = true`, and just close this PR, but I think it is nice to have these gosats logs honor the ratelimit logging config anyway. I don't have a strong opinion, though, so happy to close the PR if this is something not worth it :)

mattklein123: I would just nop these functions personally since I can't see how these logs have any value whatsoever. The rest seems fine.

irl-segfault:
what happened here? This logging produces a ton of noise and would be great to get this silence-able

PR #587, opened 2024-05-16
Add CertProvider to hot reload TLS certs for gRPC service

jayme-github:
This implements a goruntime instance to watch for on disk changes of gRPC certificates.

Certificates will be reloaded in case of change and the gRPC service will always fetch the latest one via GetCertificate on new connections instead of the static Certificates slice.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,11 @@ BUILDX_PLATFORMS := linux/amd64,linux/arm64/v8
 # Root dir returns absolute path of current directory. It has a trailing "/".
 PROJECT_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 export PROJECT_DIR
+ifneq ($(shell docker compose version 2>/dev/null),)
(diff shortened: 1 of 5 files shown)


mattklein123 (changes requested):
Can you add documentation for this change and optimally some kind of test?

jayme-github:
> Can you add documentation for this change and optimally some kind of test?

I've added documentation for it. Regarding tests I'm a bit unsure. I could write tests for `CertProvider` which would need to heavily mock the goruntime parts or tests them as well (which I don't think is desired). What would you suggest to write tests for here?

mattklein123:
> What would you suggest to write tests for here?

In a perfect world some kind of integration test.

jayme-github:
> > What would you suggest to write tests for here?
> 
> In a perfect world some kind of integration test.

Done. Hope this looks like something you would expect.

jayme-github:
@mattklein123 would you mind taking a look again please?

mattklein123 (approved):

Issue #605, opened 2024-05-27
Redis connection Error logged in plain text


jwillker:
Here https://github.com/envoyproxy/ratelimit/blob/ca55e1b31d023d8b8aedb8a21ca3b769fb632e95/src/redis/driver_impl.go#L45

When this error happens, the error is logged as plain text instead of using the Logger package.

Here and in many other places using `fmt.Println`, we should use the logger package so logs can be easily indexed in tools like Elasticsearch, Loki, or others.

The solution is to replace every Println when it should be `logger.Error()` 

Also, there is a panic here: https://github.com/envoyproxy/ratelimit/blob/ca55e1b31d023d8b8aedb8a21ca3b769fb632e95/src/redis/driver_impl.go#L69

We should use a fatal maybe so the application stops immediately with an error message.

PR #606, opened 2024-05-27
fix(redis): use logger instead of println

jwillker:
Fix: https://github.com/envoyproxy/ratelimit/issues/605

cc: @envoyproxy/ratelimit-maintainers


mattklein123 (dismissed):

mattklein123:
I think this might be a real CI issue?

jwillker:
There is a test that checks if there is panic or not; for different situations, I will maintain the panic and remove the println, so the tests will pass now.

mattklein123 (approved):

Issue #597, opened 2024-05-26
ShouldRateLimit error is never logged


jwillker:
I'm facing an error that is hard to debug because the error is never logged.

I can see in the metric `ratelimit_service_should_rate_limit_error` grow, but I cannot understand the reason of the errors.

Changing the log level to debug is possible only to see this log: https://github.com/envoyproxy/ratelimit/blob/ca55e1b31d023d8b8aedb8a21ca3b769fb632e95/src/service/ratelimit.go#L288

But the `err` itself is not logged; it is returned by the function and can only be seen by the Envoy that makes the call.
Because at the end of the day, this is a gRPC handler that returns this error to the envoy https://github.com/envoyproxy/go-control-plane/blob/v0.12.0/envoy/service/ratelimit/v3/rls.pb.go#L996

But in my case, I'm using the Ratelimit service for thousands of pods with Envoy as a sidecar(Istio), and it can be hard to find the pod that receives the error, enable the debug log, and see the error.

My purpose is to log the error in the Ratelimit service so that I can look up the error message.

I can open a PR to add the error log something like this:

```go

func (this *service) ShouldRateLimit(
	ctx context.Context,
	request *pb.RateLimitRequest) (finalResponse *pb.RateLimitResponse, finalError error) {
	.....
	defer func() {
		err := recover()
		if err == nil {
			return
		}

		logger.Debugf("caught error during call")
		logger.Error(err) // <--- New log entry
        ....
	return response, nil
}

```

PR #612, opened 2024-06-09
feat: add log error in should ratelimit

jwillker:
Fix: https://github.com/envoyproxy/ratelimit/issues/597

cc: @envoyproxy/ratelimit-maintainers

### DIFF
--- a/src/service/ratelimit.go
+++ b/src/service/ratelimit.go
@@ -285,7 +285,8 @@ func (this *service) ShouldRateLimit(
 			return
 		}
 
-		logger.Debugf("caught error during call")
+		logger.Debugf("caught error during call: %v", err)
+
 		finalResponse = nil
 		switch t := err.(type) {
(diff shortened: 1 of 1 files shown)


src/service/ratelimit.go
```diff
@@ -286,6 +286,8 @@ func (this *service) ShouldRateLimit(
 		}
 
 		logger.Debugf("caught error during call")
+
+		logger.Error(err)
```
mattklein123, line 290: This can log spam pretty badly. If we want to log something here I would have some kind of back off logger that warns periodically. Feel free to print the err as part of the debug log above.

mattklein123 (changes requested):

src/service/ratelimit.go
jwillker: For now, I just added the error to the debug.

mattklein123 (approved):

PR #648, opened 2024-07-12
Update example prom-statsd-exporter config

jcdavis:
Few cleanups for folks using this as a baseline (aka myself):

1. Remove `timer_type` for counter metrics, this field is a no-op as-is and is just confusing.
2. Where it is needed (ratelimit_service_rate_limit_shadow_mode), switch to `observer_type`, since `timer_type` has been deprecated since at least 2020: https://github.com/prometheus/statsd_exporter/commit/4a64979563442b948c6b929daee10f3b15493681
3. Add a single-key shadow mode match (even if this is not necessary/used by the example ratelimit setup)


examples/ratelimit/.DS_Store
ysawa0, line 1: remove

Issue #654, opened 2024-07-25
Security issue  (Trivy_vulnerability_report)


johnzheng1975:
https://nvd.nist.gov/vuln/detail/CVE-2024-24790

johnzheng1975:
level is critical from report :-)

talonnguyen:
hi @johnzheng1975, I'm glad seeing you are addressing the CVE. May I ask do you have the rough estimated available date? Thanks

johnzheng1975:
sorry, not got you. Available date for what?
I submitted a PR to resolve this issue, in new go version, this issue is disapppeared.

PR #656, opened 2024-07-25
improve go version, to fix security issue.

johnzheng1975:
Fix security issue:
https://github.com/envoyproxy/ratelimit/issues/654
image


kanurag94:
Does the Dockerfile here also need an update https://github.com/envoyproxy/ratelimit/blob/main/Dockerfile#L1

johnzheng1975:
Also used latest 1.22 version for docker build.
Please review, thanks.

mattklein123 (approved):

PR #664, opened 2024-08-02
chore: run gofumpt

zirain:


mattklein123 (approved):
It would be nice to do this in CI and error on it if you can do that as a follow up?

zirain:
will take a look at later

PR #663, opened 2024-07-31
implemnt Prometheus sink

zirain:
this patch try to add native prometheus support instead of a statsd_exporter sidecar container.

@mattklein123 can you take a first look, I will finish the test if this's acceptable.

Issue #285, opened 2021-08-24
Support a prometheus / openmetrics endpoint


jespersoderlund:
To be a part of an openmetrics centric metrics infrastructure it would be useful if the ratelimit service natively supported a prometheus metrics endpoint without having to go past a statsd bridge

ysawa0:
Great idea, that makes sense to me.

irl-segfault:
+1, would be super useful!  Is there any kind of shim or adapter people can use in the meantime?

irl-segfault:
^ answer: use statsd-prom exporter

zirain:
[this PR](https://github.com/lyft/gostats/pull/119) tried to fix this by adding `prometheus_sink`, but it won't.
a valid metric in stats may be invalid in prometheus, e. g. `ratelimit.service.rate_limit.httpbin.header_match_\get.near_limit`

ramaraochavali:
@zirain Do you know what needs to be done to fix ^^?

upgle:
Hi, I've abstracted an existing structure that was fully coupled to gostats and applied Prometheus metrics to it.
I would like to contribute this code. (not using gostats prometheus_sink)

I tried to contribute a small change to this project last time, but no one was interested and it closed automatically. I was wondering if there are other procedures like mailing.

Thank you

<img width="549" alt="image" src="https://user-images.githubusercontent.com/5635513/231630235-cef65f09-4c9c-4370-be5b-2913f86c4877.png">

jespersoderlund:
@upgle 
One thing that the current setup going through statsd => prometheus allows us to do is to *not* publish all metrics since some it can be quite high cardinality.

So we only publish the "near_limit" and "over_limit" metrics.

Will your implementation allow something similar, like having an allow-list of metrics to publish?

upgle:
Thank you for your kind reply.
I understand your concerns about cardinality.

I checked this out, other metrics other than "near_limit" and "over_limit" are all useful metrics (e.g. redis, memcached, localcache...), and these metrics do not have high cardinality.

Rather, a metric such as near_limit has a high cardinality because it has a descriptor key as a label for prometheus. 

And there is currently no implementation such as allow-list, but I think it is a point that can be improved.

I'll be contributing code soon and listening to feedback on slack, github, and more.

birdayz:
We could either make each metric configurable (on/off). In addition, anyone can use prometheus labeldrop to reduce cardinality. Do we need anything beyond that?

[...]

PR #681, opened 2024-08-14
feat: add Prometheus sink

zirain:
replace https://github.com/envoyproxy/ratelimit/pull/663

I tried to use the native prometheus library but failed because of complex package dependencies.
Right now most of the code came from[statsd_exporter](https://github.com/prometheus/statsd_exporter), this will allow do similar thing in a single container and without tcp request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@
   - [DogStatsD](#dogstatsd)
     - [Example](#example)
     - [Continued example:](#continued-example)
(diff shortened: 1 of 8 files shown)


zirain:
@keithmattix PTAL

arkodg:
ptal @mattklein123

jespersoderlund:
Issue #285

mattklein123 (approved):

PR #687, opened 2024-08-23
Add setting to allow runtime stats to be disabled

oscarsilver:
The runtime stats exposed by the [gostats](https://github.com/lyft/gostats) library are quite spammy and incur a lot of costs when exposed as custom metrics to DataDog. 

In this PR the option to disable the RuntimeStats has been added to make them optional.


oscarsilver:
@mattklein123 PTAL

PR #694, opened 2024-09-02
fix go-fumpt args

zirain:
follow up: https://github.com/envoyproxy/ratelimit/pull/664#pullrequestreview-2216499279


zirain:
@mattklein123 PTAL

zirain:
not stale

arkodg:
cc @collin-lee @psbrar99

PR #716, opened 2024-09-25
Sanitize stats name for remote address

zirain:
xref: https://github.com/envoyproxy/gateway/issues/4281

alternative for https://github.com/envoyproxy/envoy/pull/36275, because of what [Joshua said](https://github.com/envoyproxy/envoy/pull/36275#discussion_r1772060469).

> I feel like this change is going to spawn a lot of other changes as well. THere are a ton of places where IP addresses and other things with dots show up in stat names and I'm unsure why this one is different.

### DIFF
--- a/src/utils/utilities.go
+++ b/src/utils/utilities.go
@@ -63,12 +63,10 @@ func MaskCredentialsInUrl(url string) string {
 	return strings.Join(urls, ",")
 }
 
-var (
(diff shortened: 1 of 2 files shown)


arkodg (approved):
LGTM thanks !

arkodg:
ptal @ysawa0 @mattklein123

zirain:
not stale

Issue #244, opened 2021-04-15
Add a way to check rate limits without using quota


dweitzman:
The grpc [RateLimitService](https://github.com/envoyproxy/envoy/blob/main/api/envoy/service/ratelimit/v3/rls.proto) has only one method `ShouldRateLimit` and `hits_addend` is assumed to be 1 when no value has been explicitly set.

**Why would a person want to get quota without incrementing it?**

One example is if you're dealing with expensive requests but you don't know how expensive they are until _after_ they've run. Maybe you add up the CPU seconds spent serving a request an increment it at the end of the request. You'd still need to check (but not increment) at the beginning of requests whether the CPU-second quota has been exhausted by previous requests during the current time unit.

A hacky workaround today might be to make sure that expensive requests are measured in large numbers (like hundreds of thousands) and check quota by incrementing a very small number (like 1) that might effectively act as if you haven't incremented it at all.

**How might this be implemented?**

Seems like there are a few possible approaches to getting quota without changing it:
- Do something clever and subtle like saying there's a `hits_addend` of uint32_max means return the quota instead of incrementing it
- Add some kind of a `check_only` flag to the `RateLimitRequest` message
- Make a separate `CheckRateLimit` method on `RateLimitService`
- Make a separate non-standard grpc service specific to this rate limiter implementation so the `RateLimitService` in envoy proxy doesn't need to change

mattklein123:
I don't feel strongly about it, but `check_only` sounds reasonable to me.

dweitzman:
> Is there any update on getting this feature added?

Seems like the maintainers are open to reviewing code, if you're open to implementing it.

I filed this issue to check in on what design approach would most easily get through code review. It was never a high enough priority for me to implement before I left Pinterest. Maybe you're the new me and the priority is higher now 🙂

brightbyte:
From my (naive and limited) understanding,  `hits_addend = 0` should work for this... doesn't it? That's how I have seen this feature implemented in other rate limiting systems.

PR #729, opened 2024-10-09
Make default hitsAddend minimum value configurable

gcalmettes:
fix: https://github.com/envoyproxy/ratelimit/issues/730

In the current implementation of the `ratelimit` service, the `hitsAddend` value that is associated with each request is set to a minimum of `1`.
This is done to ensure that if the `RateLimit` request does not specify any `request.hitsAddend` (like what the envoy ratelimit filter does by default if no [envoy.ratelimit.hits_addend](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/ratelimit/v3/rls.proto#envoy-v3-api-field-service-ratelimit-v3-ratelimitrequest-hits-addend) has been defined), the hit is still counted.

This minimum `hitsAddend` of `1` is enforced via a [`Max(1, <hitsAddend from request>)`](https://github.com/envoyproxy/ratelimit/blob/28b1629a21e885bdd2b527d6a1c1de8483dc47d4/src/redis/fixed_cache_impl.go#L52) function.

This PR adds the possibility to configure the minimum value of the `hitsAddend` via a settings, so `0` can also be used as minimum `hitsAddend` value if needed. The current behavior of `1` being the default value is preserved, and if the environment variable is not set, then a value of `1` is applied, like currently.

The ability to configure a minimum `hitsAddend` to zero can help for example for the following use cases:
- having the possibility to send a `RateLimitRequest` to check the current remaining rates without consuming a quota (e.g.: https://github.com/envoyproxy/ratelimit/issues/244)
- supporting the cases when the "weight" of the request is not known during the request but on the response (e.g.: rate limiting llm tokens as per https://github.com/envoyproxy/envoy/issues/29161#issuecomment-1867463631). In this case, we still want to ratelimit on the request, but we do not want the request itself to add an additional arbitrary `1` value to the real weight defined as `hitsAddend`.


gcalmettes:
@envoyproxy/ratelimit-maintainers

zirain:
I think @mattklein123 is the person, hope he have some free slot to review.

gcalmettes:
Any chance you could take a look @mattklein123 ? Thanks a lot.

gcalmettes:
commenting to prevent automatic closing by stale-bot

src/settings/settings.go
```diff
@@ -106,6 +106,7 @@ type Settings struct {
 	CacheKeyPrefix                     string  `envconfig:"CACHE_KEY_PREFIX" default:""`
 	BackendType                        string  `envconfig:"BACKEND_TYPE" default:"redis"`
 	StopCacheKeyIncrementWhenOverlimit bool    `envconfig:"STOP_CACHE_KEY_INCREMENT_WHEN_OVERLIMIT" default:"false"`
+	HitsAddendMinValue                 uint32  `envconfig:"DEFAULT_HITS_ADDEND_MIN_VALUE" default:"1"`
```
arkodg, line 109: can we do w/o this extra field ?
we'll need to make HitsAddend optional in the API https://github.com/envoyproxy/go-control-plane/blob/e0e383246f4cd59a40967987057c76f03526cffb/envoy/service/ratelimit/v3/rls.pb.go#L179
which may be a breaking change for clients
cc @zirain @mathetake @wbpcode

mathetake: if this still defaults to 1 then i don't think this would be a breaking change in practice i guess? but ack we should need to allow zero in the API regardless

wbpcode:
What's is the behavior of zero hits_addend? I think we use 1 as check and CUSTOM_HITS_ADDEND -1 as report, but if 0 could aslo be used as check, maybe it's simpler.

src/settings/settings.go
gcalmettes: @arkodg I confirm that this is not a breaking change and that the `HitsAddendMinValue` field introduced here is not needed in the `RatelimitRequest` request send to the ratelimit service.

This is only a config variable of the reatelimit service, with default value of 1 that matches the current **hardcoded** value of 1.

@mathetake there is no change to be done to the API, as zero is already a valid value for the `HitsAddend` value. In fact it is the **default value** when no custom `HitsAddend` value is set via the [Set-Filter-State](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/set_filter_state) HTTP filter or directly in the `RatelimitRequest` and the actual `+1` increment is enforced at the ratelimit service level (if the hitAddend value is set to zero in the request -- either voluntarily or because it is absent in the request-- the value is [override to `1`](https://github.com/envoyproxy/ratelimit/blob/28b1629a21e885bdd2b527d6a1c1de8483dc47d4/src/redis/fixed_cache_impl.go#L52). 

Currently, if someone uses the  `FilterState` to enforce the hitsAddend value **voluntarily** to zero (we have a use case for that) or set the `HitsAddend` value to zero in the `RatelimitRequest`, then in practice the limit counter will still be incremented because of the `max(request.HitsAddend, 1)` check that is done, which is not expected. 

What this PR does basically is just to allow to change this hardcoded `1` to any value `>=0` ([same constraints than the api](https://github.com/EItanya/envoy/blob/b1a2af6a29b6a6d95449d407c9e3e5ae64c74788/api/envoy/service/ratelimit/v3/rls.proto#L50), but still keeping 1 as default value to not change the actual behavior) to allow `RatelimitRequests` with zero increment.

mathetake:
So currently, [RateLimitRequest](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/ratelimit/v3/rls.proto) API assumes that null addend should be interpreted as 1: 

> (uint32) Rate limit requests can optionally specify the number of hits a request adds to the matched limit. If the value is not set in the message, a request increases the matched limit by 1. 

so if we change the default value to whatever number to non one, then RSL doesn't comply with the API comment in other words. In fact, Envoy's current implementation relies on the fact that "`zero` must be interpreted as 1":

https://github.com/envoyproxy/envoy/blob/ab9ff8e5aa5735d3bb796bd27c44638c292c5593/source/extensions/filters/http/ratelimit/ratelimit.cc#L92-L95

But I think we could make changes to Envoy as well that 
* remove the comment of defaulting to 1
* explicitly use 1 as a default value in the implementation

wdyt

mathetake:
> I would suggest to use per-descriptor hits_addend to replace the per-request hits_addend. Note that every descriptor is evaluated independent anyway. The new per-descriptor hits_addend is more flexible and has no history burden.

sounds good, is that the feature-to-be-added in https://github.com/envoyproxy/envoy/pull/37567?

I think then the per-descriptor addend combined with this configuration makes our life easier

src/settings/settings.go
arkodg: hey @gcalmettes, we understand the issue and our use case is the same as yours :)
I was questioning the current RLS API and the semantics defined in [RateLimitRequest](https://github.com/envoyproxy/go-control-plane/blob/e0e383246f4cd59a40967987057c76f03526cffb/envoy/service/ratelimit/v3/rls.pb.go#L163C6-L163C22) , since its not a pointer, there's no way to figure out unset (default:1 ) vs set to 0.
It maybe too late to edit that API, the easiest thing to do maybe is 
* set the default value to `1` in envoy if `HitsAddend` is not set
* update the code in this project to `max(request.HitsAddend, 0)` simplifying the logic and maintaining backwards compatibility
wdyt @mathetake @wbpcode

arkodg: sounds like you had the same idea @mathetake https://github.com/envoyproxy/ratelimit/pull/729#issuecomment-2539365446

wbpcode:
> > I would suggest to use per-descriptor hits_addend to replace the per-request hits_addend. Note that every descriptor is evaluated independent anyway. The new per-descriptor hits_addend is more flexible and has no history burden.
> 
> sounds good, is that the feature-to-be-added in [envoyproxy/envoy#37567](https://github.com/envoyproxy/envoy/pull/37567)?
> 
> I think then the per-descriptor addend combined with this configuration makes our life easier

Yeah. The #37567 is the first PR to add API and local rate limit implementation. Then I will add a PR to support it for global ratelimit. At that PR is #37567 is merged, I can do the per-descriptor support the `envoyproxy/ratelimit` as same time.

lizzzcai:
> Yeah. The #37567 is the first PR to add API and local rate limit implementation. Then I will add a PR to support it for global ratelimit. At that PR is #37567 is merged, we can do the per-descriptor support the envoyproxy/ratelimit as same time.

Hi @wbpcode, is there a plan to support per-descriptor feature you mentioned above in the ratelimit service. it will be a good feature to ratelimit both rps and token. Thanks.

lizzzcai:
Sorry, just realized it was already implemented in #802, amazing, thanks!

PR #752, opened 2024-11-07
support ReturnDescriptorsInResponse

zirain:
this patch will allow us save descriptors through [envoy built-in mechanism](https://github.com/envoyproxy/envoy/blob/1c90708283e6514407d66ca78a20398738e71514/source/extensions/filters/http/ratelimit/ratelimit.cc#L179), it will be help to some additional check(e.g. aduit, post-ratelimit check) in other filters.

### DIFF
--- a/src/service/ratelimit.go
+++ b/src/service/ratelimit.go
@@ -1,25 +1,21 @@
 package ratelimit
 
 import (
+	"encoding/json"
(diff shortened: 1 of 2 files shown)


zirain:
replaced by https://github.com/envoyproxy/envoy/pull/37548

Issue #764, opened 2024-11-14
huge gostats logs by default.


johnzheng1975:
I just installed ratelimit by default, however, the huge logs raise for go stats -- 177M / hour.
Somehow it is just for debug, we had better set DISABLE_STATS to true by default. Otherwise, may bring confuse to user.

Thanks for consideration.
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: ratelimit
  namespace: ratelimit
spec:
  replicas: ${RATELIMIT_REPLICAS}
  selector:
    matchLabels:
      app: ratelimit
  strategy:
    type: Recreate
  template:
    metadata:
      labels:
        app: ratelimit
    spec:
      containers:
      - image: envoyproxy/ratelimit:30a4ce1a
        imagePullPolicy: Always
        name: ratelimit
        command: ["/bin/ratelimit"]
        env:
        - name: LOG_LEVEL
          value: info
        - name: REDIS_SOCKET_TYPE
          value: tcp
        - name: REDIS_URL
          value: redis:6379
        - name: USE_STATSD
          value: "false"
        - name: RUNTIME_ROOT
          value: /data
        - name: RUNTIME_SUBDIRECTORY
          value: ratelimit
        - name: RUNTIME_WATCH_ROOT
          value: "false"
        - name: RUNTIME_IGNOREDOTFILES
          value: "true"
        - name: HOST
          value: "::"
        - name: GRPC_HOST
          value: "::"
        ports:
        - containerPort: 8080
        - containerPort: 8081
        - containerPort: 6070
        resources:
          limits:
            cpu: ${RATELIMIT_LIMIT_CPU}
            memory: ${RATELIMIT_LIMIT_MEMORY}
          requests:
            cpu: ${RATELIMIT_REQUEST_CPU}
            memory: ${RATELIMIT_REQUEST_MEMORY}
        volumeMounts:
        - name: config-volume
          mountPath: /data/ratelimit/config
      volumes:
      - name: config-volume
        configMap:
          name: ratelimit-config
```

johnzheng1975:
This is huge logs:
```
...
{"level":"debug","ts":1731575723.674967,"logger":"gostats.loggingsink","msg":"flushing timer","json":{"name":"ratelimit_server.ShouldRateLimit.response_time","type":"timer","value":"0.000000"}}
[...]

PR #765, opened 2024-11-14
set envconfig DISABLE_STATS true by default

johnzheng1975:
Fix issue: https://github.com/envoyproxy/ratelimit/issues/764


zirain:
I believe it's better to keep the default as before, it's easy for people to change it in real/prod env.

johnzheng1975:
That's fine.  Close it.

PR #774, opened 2024-12-01
swith to github.com/libp2p/go-reuseport

zirain:
alternative for https://github.com/envoyproxy/ratelimit/pull/748


arkodg:
the motivation (my guess is) for using the `reuseport` package is to set `SO_REUSEADDR` & `SO_REUSEPORT` e.g.
https://github.com/libp2p/go-reuseport/blob/a3a76c79205aca3ed2bfd07c6c45f29e8cb0784d/control_unix.go

Since the existing package doesnt work well for binding to an ipv6 addr, my suggestion is to replace it with https://github.com/libp2p/go-reuseport or setup the listenConfig in this repo itself

cc @mattklein123

arkodg (approved):
LGTM thanks !

psbrar99 (approved):
looks good. Thanks.

Issue #796, opened 2024-12-17
ResourceExhausted while trying to send message larger config


renuka-fernando:
Getting the following error when the xDS server sends a larger config.

level=error msg="Failed to receive configuration from DS Management Server: pc error: code = ResourceExhausted desc = grpc: received message larger than max (4197803 vs. 4194304)*

PR #797, opened 2024-12-17
Add configuration XDS_CLIENT_MAX_MSG_SIZE_IN_BYTES

renuka-fernando:
### Description
Fix https://github.com/envoyproxy/ratelimit/issues/796

### Tasks
- [x] Doc update for configs

### DIFF
--- a/src/settings/settings.go
+++ b/src/settings/settings.go
@@ -78,8 +78,8 @@ type Settings struct {
 	XdsClientBackoffRandomFactor    float64       `envconfig:"XDS_CLIENT_BACKOFF_RANDOM_FACTOR" default:"0.5"`
 	XdsClientBackoffJitter          bool          `envconfig:"XDS_CLIENT_BACKOFF_JITTER" default:"true"`
 
-	// xDS gRPC client options
-	XdsGrpcClientOptionsMaxMsgSizeInBytes int `envconfig:"XDS_GRPC_CLIENT_OPTIONS_MAX_MSG_SIZE_IN_BYTES" default:""`
+	// xDS client gRPC options
+	XdsClientGrpcOptionsMaxMsgSizeInBytes int `envconfig:"XDS_CLIENT_MAX_MSG_SIZE_IN_BYTES" default:""`
 
 	// Stats-related settings
 	UseDogStatsd           bool              `envconfig:"USE_DOG_STATSD" default:"false"`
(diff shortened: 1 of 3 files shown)


renuka-fernando:
hi @collin-lee, @arkodg, @psbrar99

Could you please have a look?

arkodg:
can you fix the title @renuka-fernando

src/settings/settings.go
```diff
@@ -78,6 +78,9 @@ type Settings struct {
 	XdsClientBackoffRandomFactor    float64       `envconfig:"XDS_CLIENT_BACKOFF_RANDOM_FACTOR" default:"0.5"`
 	XdsClientBackoffJitter          bool          `envconfig:"XDS_CLIENT_BACKOFF_JITTER" default:"true"`
 
+	// xDS gRPC client options
+	XdsGrpcClientOptionsMaxMsgSizeInBytes int `envconfig:"XDS_GRPC_CLIENT_OPTIONS_MAX_MSG_SIZE_IN_BYTES" default:""`
```
arkodg, line 82: 1. the previous env vars dont have the `grpc` prefix
2. should we separate send and receive ? do we even to add a knob for send, its likely to be small

renuka-fernando: @arkodg thanks for your review.
1. changed the config as `XDS_CLIENT_GRPC_OPTIONS_MAX_MSG_SIZE_IN_BYTES `
2. Yes, since [discoveryrequest](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/discovery/v3/discovery.proto#envoy-v3-api-msg-service-discovery-v3-discoveryrequest) is smaller removed the send message size.

arkodg: should this be `XDS_CLIENT_MAX_MSG_SIZE_IN_BYTES`, since the previous grpc options didnt contain the `grpc` prefix ?

renuka-fernando: Updated via commit: https://github.com/envoyproxy/ratelimit/pull/797/commits/39135a7bc9daa4779e08702ee59939dec4480f18

arkodg (approved):
LGTM thanks !

src/settings/settings.go
arkodg: looks good, thanks

collin-lee (approved):

PR #802, opened 2024-12-23
ratelimit: per descriptor hits addend support and prefer uint64

wbpcode:


mathetake (commented):
high level question: if the number zero is given, it should work like "check the budget is not negative, and if not requests are allowed"? I am asking because if this would obviate the issue we had in https://github.com/envoyproxy/ratelimit/pull/729

wbpcode:
> high level question: if the number zero is given, it should work like "check the budget is not negative, and if not requests are allowed"? I am asking because if this would obviate the issue we had in #729

The counter in redis is a number that counts all used tokens. So, we only need to check if the counter is larger than the threshold or not.

mathetake:
cool, so in other words, sending hits_addend=0 can be used to reject requests if the counter in redis is negative or zero ya?

wbpcode:
> cool, so in other words, sending hits_addend=0 can be used to reject requests if the counter in redis is negative or zero ya?

sorry, I may didn't make it clear. The counter will never be negative in the redis. It is a counter the will increases forever in the time window.

But the conclusion is right. hits_addend zero could also could be used to reject a request.

mathetake:
great, thank you for clarifying!

mathetake (approved):

src/redis/fixed_cache_impl.go
```diff
@@ -32,119 +32,131 @@ type fixedRateLimitCacheImpl struct {
...
 	*pipeline = client.PipeAppend(*pipeline, result, "INCRBY", key, hitsAddend)
 	*pipeline = client.PipeAppend(*pipeline, nil, "EXPIRE", key, expirationSeconds)
 }
 
-func pipelineAppendtoGet(client Client, pipeline *Pipeline, key string, result *uint32) {
+func pipelineAppendtoGet(client Client, pipeline *Pipeline, key string, result *uint64) {
 	*pipeline = client.PipeAppend(*pipeline, result, "GET", key)
 }
 
+func (this *fixedRateLimitCacheImpl) getHitsAddend(hitsAddend uint64, isCacheKeyOverlimit, isCacheKeyNearlimit,
```
mathetake, line 44: do you want to add unit tests for this function?

wbpcode: Of cource. I will add some unit tests for the whole feature.

wbpcode:
cc @psbrar99

mathetake:
also @zirain if you have any comments

src/utils/utilities.go
```diff
@@ -41,7 +41,7 @@ func CalculateReset(unit *pb.RateLimitResponse_RateLimit_Unit, timeSource TimeSo
 	return &durationpb.Duration{Seconds: sec - now%sec}
 }
 
-func Max(a uint32, b uint32) uint32 {
+func Max(a uint64, b uint64) uint64 {
```
zirain, line 44: not a must, but is it find to remove this as golang support generic?

zirain (approved):
LGTM, only a small nit.

src/utils/utilities.go
zirain:
```
// The max built-in function returns the largest value of a fixed number of
// arguments of [cmp.Ordered] types. There must be at least one argument.
// If T is a floating-point type and any of the arguments are NaNs,
// max will return NaN.
func max[T cmp.Ordered](x T, y ...T) T
```

psbrar99:
LGTM, thanks @wbpcode .

psbrar99 (approved):

mathetake:
fantastic, thank you for your hard work! @wbpcode 👯

PR #837, opened 2025-02-12
Update Example 2 with details about STOP_CACHE_KEY_INCREMENT_WHEN_OVERLIMIT

taer:
Updating README about something I missed when setting up multiple descriptor rate limit.


taer:
Just a README update. helps others know how an option relates to the config

taer:
@collin-lee @arkodg    i saw y'all recently looked at PRs. This should be a simple one

PR #922, opened 2025-07-15
Update codeql with fetch depth 0

collin-lee:
Allow CodeQL to compute diffs properly and resolve undefined errors in restrictAlertsTo.


collin-lee:
@dependabot rebase

Issue #437, opened 2023-07-21
provide distroless image


infa-ddeore:
since its just a single go binary, can we have distroless image instead of alpine?

infa-ddeore:
no updates?

Issue #396, opened 2023-02-16
Discussion: Ideal base image to build the container image


AmitKumarDas:
I notice that `ratelimit` makes use of alpine as the base image to build its final image. I was wondering if there were any discussions to support other distros esp. distroless (which is based on Debian). Alternatively, is the current ratelimit binary dependent on Alpine?

PR #985, opened 2025-10-22
Use a scratch based image to limit vulnerabilities

morepork:
The alpine base image isn't needed as the statically linked go binary runs fine without it. Using a scratch based image reduces the number of vulnerabilities brought up by scanning tools in the alpine image, and makes the image a bit smaller.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,9 @@ COPY script script
 
 RUN CGO_ENABLED=0 GOOS=linux go build -o /go/bin/ratelimit -ldflags="-w -s" -v github.com/envoyproxy/ratelimit/src/service_cmd
 
-FROM alpine:3.22.2@sha256:4b7ce07002c69e8f3d704a9c5d6fd3053be500b7f1c69fc0d80990c2ad8dd412 AS final
+FROM alpine:3.22.2@sha256:4b7ce07002c69e8f3d704a9c5d6fd3053be500b7f1c69fc0d80990c2ad8dd412 AS cacerts
 RUN apk --no-cache add ca-certificates && apk --no-cache update
+
+FROM scratch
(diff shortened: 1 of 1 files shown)


collin-lee:
Are we just using Alpine image for the certs?  I see the change considers that.  I think this is okay @arkodg

arkodg:
hey there seem to be multiple GH issues around this
* https://github.com/envoyproxy/ratelimit/issues/437
* https://github.com/envoyproxy/ratelimit/issues/396

my recommendation would be to add a new distroless variant, if this approach is taken, the decision to be made is
* publish two images - one with a `distroless-` prefix in the tag and one without ( alpine)
* or publish one distroless image and rm alpine but that may break downstream users who have been relying on using the shell for debugging etc

morepork:
My preference would be to just switch to a distroless image. It would still be possible to debug via an ephemeral debug container.

Having 2 images would also be fine, we could go with `<tag>` and `distroless-<tag>` as Arko suggests, which would be [consistent with Envoy](https://www.envoyproxy.io/docs/envoy/latest/start/install#start-install-docker), but in my experience that leads to a lot of people using the non-distroless variant when they should really be using the distroless one, I always prefer to see a secure by default approach. Another naming scheme I've seen is to have `<tag>` be distroless, and also having a `<tag>-debug` or `<tag>-alpine`.

For us any option is ok, I'm happy to go with whatever you prefer.

arkodg:
thanks and agree @morepork, just switching to `distroless` should be _okay_

collin-lee:
like the distroless approach

collin-lee:
Maybe something like this? 

https://github.com/envoyproxy/ratelimit/pull/993

morepork:
Closing in favour of https://github.com/envoyproxy/ratelimit/pull/993

PR #993, opened 2025-10-27
Replace Alpine with Google's distroless static image for enhanced sec…

collin-lee:
…urity

and simplified maintenance. Includes CA certificates automatically and provides debug variant for troubleshooting.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,5 +10,5 @@ COPY script script
 
 RUN CGO_ENABLED=0 GOOS=linux go build -o /go/bin/ratelimit -ldflags="-w -s" -v github.com/envoyproxy/ratelimit/src/service_cmd
 
-FROM gcr.io/distroless/static-debian12
+FROM gcr.io/distroless/static-debian12:nonroot@sha256:e8a4044e0b4ae4257efa45fc026c0bc30ad320d43bd4c1a7d5271bd241e386d0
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -10,6 +10,5 @@ COPY script script
 
 RUN CGO_ENABLED=0 GOOS=linux go build -o /go/bin/ratelimit -ldflags="-w -s" -v github.com/envoyproxy/ratelimit/src/service_cmd
 
-FROM alpine:3.22.2@sha256:4b7ce07002c69e8f3d704a9c5d6fd3053be500b7f1c69fc0d80990c2ad8dd412 AS final
-RUN apk --no-cache add ca-certificates && apk --no-cache update
+FROM gcr.io/distroless/static-debian12
```
morepork: what do you think about pinning this to a particular SHA to make builds deterministic?

Also, I think it may as well use the `nonroot` tag, similar to the envoy build:
https://github.com/envoyproxy/envoy/blob/main/distribution/docker/Dockerfile-envoy#L62

collin-lee:
@arkodg

Dockerfile
collin-lee: @morepork see https://github.com/envoyproxy/ratelimit/pull/993/commits/5009d932c9b8e200d4bdd2e3a176bb3f4b0d20ee

Issue #994, opened 2025-10-30
[Proposal] Include value in metrics for unspecified value


xuannam230201:
Currently, `envoyproxy/ratelimit` support `detailed_metric` to include all values of the descriptor trees into metrics. But this causes high cardinality in metrics in some cases.

For example:

```
domain: default
descriptors:
  - key: route
    descriptors:
      - key: http_method
        descriptors:
          - key: user_id
            detailed_metric: true
            rate_limit:
              unit: minute
              requests_per_unit: 60
```

This rate limiting rule indicates per endpoint per user id rate limits. With `detailed_metric: true`, metric includes route, http_method and subject_id's values which cause high cardinality (num of endpoints * num of users).

Is it possible to add a field `value_to_metric: true` to decide to include which values to metric? Or we need to implement this feature?

In this case, I expect that I only need to include route and key's values to metric (the number of labels would be num of endpoints), I can configure:
```
domain: default
descriptors:
  - key: route
    value_to_metric: true
    descriptors:
      - key: http_method
        value_to_metric: true
        descriptors:
          - key: user_id
            rate_limit:
              unit: minute
              requests_per_unit: 60
```

xuannam230201:
hi @collin-lee , @psbrar99  could you please help me review/support this feature request? OR should I create a PR for that?

xuannam230201:
Hi @collin-lee, @psbrar99 , @arkodg  , could you please help to review this PR: https://github.com/envoyproxy/ratelimit/pull/996 
Thank you!

xuannam230201:
Hi @collin-lee , @arkodg , @psbrar99 , @mattklein123  could you please help to take a look at this 🙏 OR Please let me know if I need to contact anyone else on this. Thank you!

brightbyte:
I just came here to report the same issue,. This feature would be very useful for catch-all limits.

I it may not even be necessary to introduce a new key like `detailed_metric` with new semantics. It would be sufficient if `detailed_metric` would work at any  intermediate level, assuming that descriptor values with high cardinality are always lower in the hierarchy.

Am am thinking of a use case like this:
```yaml
domain: acme
descriptors:
      - key: user_class
        # No value specified, catch-all
        detailed_metric: true # include the class name in the metric
        descriptors:
          - key: user_id
            # Don't include the user class in the limit
            rate_limit:
              requests_per_unit: 100
              unit: MINUTE
       
      - key: user_class
        value: premium # value specified
        descriptors:
[...]

PR #996, opened 2025-10-31
feat: Add field to add unspecified value to metric

xuannam230201:
# Add `value_to_metric` field to include descriptor values in metrics

## Summary

This PR adds a new optional field `value_to_metric` (default: `false`) to each descriptor in the rate limit configuration. When enabled, it includes the descriptor's runtime value in the metric key, even when the descriptor value is not explicitly defined in the configuration. This provides visibility into different rate limit scenarios without needing to pre-define every possible value.

## Problem

Previously, when a descriptor matched a value that wasn't explicitly defined in the configuration (i.e., matched via a default key without value), the metric key would only include the descriptor key, not the actual runtime value. This made it difficult to track and analyze rate limiting metrics for different runtime values without using `detailed_metric`, which includes values for all descriptors and can lead to high cardinality.

## Solution

The new `value_to_metric` field allows users to selectively include runtime values in metric keys for specific descriptors, providing granular control over metric cardinality while still maintaining visibility into important descriptor values.

## Behavior

- **Default behavior**: When `value_to_metric` is `false` (default) or not set, the behavior remains unchanged - descriptors matched via default keys only include the key name in metrics.

- **With `value_to_metric: true`**: When enabled on a descriptor:
  - If the descriptor matches via a default key (no explicit value in config), the runtime value is included in the metric key: `domain.key_value.subkey`
  - If the descriptor matches via an explicit key+value or wildcard, the runtime value is always included in the metric key
  - When combined with wildcard matching, the **full runtime value** is included, not just the wildcard prefix

- **Precedence**: When `detailed_metric: true` is set on a descriptor, it takes precedence and `value_to_metric` is ignored for that descriptor (to maintain backward compatibility).

## Example

**Configuration:**
```yaml
domain: domain
descriptors:
  - key: route
    value_to_metric: true
    descriptors:
      - key: http_method
        value_to_metric: true
        descriptors:
          - key: subject_id
            rate_limit:
              unit: minute
              requests_per_unit: 60
```

**Requests:**
- `route=api`, `http_method=GET`, `subject_id=123` → Metric: `domain.route_api.http_method_GET.subject_id`
- `route=web`, `http_method=POST`, `subject_id=456` → Metric: `domain.route_web.http_method_POST.subject_id`

**Without `value_to_metric`**, both requests would use: `domain.route.http_method.subject_id`

### Code Changes
- Added `ValueToMetric bool` field to `YamlDescriptor` struct
- Added `value_to_metric` to `validKeys` map for YAML validation
- Added `valueToMetric bool` field to `rateLimitDescriptor` struct to track the flag per descriptor
- Updated `loadDescriptors` to store the `value_to_metric` flag in descriptor nodes
- Updated `GetLimit` to build a `value_to_metric`-enhanced metric key when enabled
- Handled wildcard matching to include full runtime values when `value_to_metric` is enabled

### Tests
- Added comprehensive unit tests covering:
  - Basic functionality with runtime values
  - Default key behavior with `value_to_metric`
  - Mid-level descriptor with `value_to_metric`
  - Backward compatibility (no flag set)
  - Interaction with `detailed_metric` (precedence)
  - Configured descriptor values with `value_to_metric`
  - Wildcard matching with `value_to_metric`
- All tests pass successfully

### Documentation
- Updated README.md with:
  - Added `value_to_metric` to descriptor list definition format
  - New section "Including descriptor values in metrics" explaining the feature
  - Example 10 demonstrating usage with basic and wildcard scenarios
  - Updated Table of Contents (note: requires running `doctoc` to regenerate)

## Testing

All existing tests continue to pass, ensuring backward compatibility. New tests verify:
- ✅ Basic `value_to_metric` functionality
- ✅ Default key behavior includes values when enabled
- ✅ Wildcard matching includes full runtime values
- ✅ No regression when flag is not set
- ✅ Correct precedence with `detailed_metric`
- ✅ Works with configured descriptor values

## Backward Compatibility

This change is fully backward compatible:
- Default value is `false`, so existing configurations continue to work unchanged
- Only affects metrics keys when explicitly enabled
- Does not change rate limiting behavior, only metric naming

### DIFF
--- a/src/config/config_impl.go
+++ b/src/config/config_impl.go
@@ -356,22 +356,28 @@ func (this *rateLimitConfigImpl) GetLimit(
 			if matchedViaWildcard {
 				if nextDescriptor.valueToMetric {
 					valueToMetricFullKey.WriteString(entry.Key)
-					valueToMetricFullKey.WriteString("_")
-					valueToMetricFullKey.WriteString(entry.Value)
+					if entry.Value != "" {
+						valueToMetricFullKey.WriteString("_")
+						valueToMetricFullKey.WriteString(entry.Value)
+					}
 				} else {
 					valueToMetricFullKey.WriteString(entry.Key)
 				}
 			} else if matchedUsingValue {
 				// Matched explicit key+value in config
 				valueToMetricFullKey.WriteString(entry.Key)
-				valueToMetricFullKey.WriteString("_")
-				valueToMetricFullKey.WriteString(entry.Value)
+				if entry.Value != "" {
+					valueToMetricFullKey.WriteString("_")
+					valueToMetricFullKey.WriteString(entry.Value)
+				}
 			} else {
 				// Matched default key (no value) in config
 				if nextDescriptor.valueToMetric {
 					valueToMetricFullKey.WriteString(entry.Key)
-					valueToMetricFullKey.WriteString("_")
-					valueToMetricFullKey.WriteString(entry.Value)
+					if entry.Value != "" {
+						valueToMetricFullKey.WriteString("_")
+						valueToMetricFullKey.WriteString(entry.Value)
+					}
 				} else {
 					valueToMetricFullKey.WriteString(entry.Key)
 				}
@@ -406,7 +412,9 @@ func (this *rateLimitConfigImpl) GetLimit(
 
 	// Replace metric with detailed metric, if leaf descriptor is detailed.
 	if rateLimit != nil && rateLimit.DetailedMetric {
-		rateLimit.Stats = this.statsManager.NewStats(detailedMetricFullKey.String())
+		detailedKey := detailedMetricFullKey.String()
+		rateLimit.Stats = this.statsManager.NewStats(detailedKey)
+		rateLimit.FullKey = detailedKey
 	}
 
 	// If not using detailed metric, but any value_to_metric path produced a different key,
@@ -417,6 +425,7 @@ func (this *rateLimitConfigImpl) GetLimit(
 			// Recreate to ensure a clean stats struct, then set to enhanced stats
 			rateLimit = NewRateLimit(rateLimit.Limit.RequestsPerUnit, rateLimit.Limit.Unit, this.statsManager.NewStats(rateLimit.FullKey), rateLimit.Unlimited, rateLimit.ShadowMode, rateLimit.Name, rateLimit.Replaces, rateLimit.DetailedMetric)
 			rateLimit.Stats = this.statsManager.NewStats(enhancedKey)
+			rateLimit.FullKey = enhancedKey
 		}
 	}
 
--- a/test/config/config_test.go
+++ b/test/config/config_test.go
@@ -1353,3 +1353,242 @@ func TestValueToMetric_WithWildcard(t *testing.T) {
 	rl.Stats.TotalHits.Inc()
 	asrt.EqualValues(1, store.NewCounter(expectedKey+".total_hits").Value())
 }
+
+func TestValueToMetric_WithEmptyValue(t *testing.T) {
+	asrt := assert.New(t)
+	store := stats.NewStore(stats.NewNullSink(), false)
+
+	cfg := []config.RateLimitConfigToLoad{
+		{
+			Name: "inline",
+			ConfigYaml: &config.YamlRoot{
+				Domain: "domain",
+				Descriptors: []config.YamlDescriptor{
+					{
+						Key:           "route",
+						ValueToMetric: true,
+						Descriptors: []config.YamlDescriptor{
+							{
+								Key:           "http_method",
+								ValueToMetric: true,
+								Descriptors: []config.YamlDescriptor{
+									{
+										Key: "subject_id",
+										RateLimit: &config.YamlRateLimit{
+											RequestsPerUnit: 60,
+											Unit:            "minute",
+										},
+									},
+								},
+							},
+						},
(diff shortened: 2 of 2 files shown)


src/config/config_impl.go
```diff
@@ -323,20 +330,55 @@ func (this *rateLimitConfigImpl) GetLimit(
...
+			matchedUsingValue = false
+		}
+
+		// Build value_to_metric metrics path for this level
+		valueToMetricFullKey.WriteString(".")
+		if nextDescriptor != nil {
+			if matchedViaWildcard {
+				if nextDescriptor.valueToMetric {
+					valueToMetricFullKey.WriteString(entry.Key)
+					valueToMetricFullKey.WriteString("_")
```
collin-lee: could entry.Value here be empty potentially?  If so, maybe add a guard

if entry.Value != "" {
    valueToMetricFullKey.WriteString("_")
    valueToMetricFullKey.WriteString(entry.Value)
}

xuannam230201:
Updated PR because of failed format checks: https://github.com/envoyproxy/ratelimit/actions/runs/19188015447/job/55114797917?pr=996.

xuannam230201:
Hi @collin-lee , do you have any concerns/feedbacks on this PR (for this issue https://github.com/envoyproxy/ratelimit/issues/994)? Is it good to go?

collin-lee:
> Hi @collin-lee , do you have any concerns/feedbacks on this PR (for this issue #994)? Is it good to go?

See comments above regarding FullKey

xuannam230201:
> See comments above regarding FullKey

@collin-lee , I don't see any comments from you regarding FullKey. Could you please let me know what are your concerns? Thanks.

image

collin-lee:
> > See comments above regarding FullKey
> 
> @collin-lee , I don't see any comments from you regarding FullKey. Could you please let me know what are your concerns? Thanks.
> 
> image

Weird... a few weeks ago I had made a comment and then the other day, I wrote this.  It looks like my comments are pending so maybe that's why you don't see it.

Looks like I didn't click on options in the dropdown to "publish" it.  do you see it now?

src/config/config_impl.go
xuannam230201: Agreed! Let me add the guard and also unit test for this.

src/config/config_impl.go
```diff
@@ -367,6 +409,17 @@ func (this *rateLimitConfigImpl) GetLimit(
...
 
+	// If not using detailed metric, but any value_to_metric path produced a different key,
+	// override stats to use the value_to_metric-enhanced key
+	if rateLimit != nil && !rateLimit.DetailedMetric {
+		enhancedKey := valueToMetricFullKey.String()
+		if enhancedKey != rateLimit.FullKey {
+			// Recreate to ensure a clean stats struct, then set to enhanced stats
+			rateLimit = NewRateLimit(rateLimit.Limit.RequestsPerUnit, rateLimit.Limit.Unit, this.statsManager.NewStats(rateLimit.FullKey), rateLimit.Unlimited, rateLimit.ShadowMode, rateLimit.Name, rateLimit.Replaces, rateLimit.DetailedMetric)
+			rateLimit.Stats = this.statsManager.NewStats(enhancedKey)
+		}
```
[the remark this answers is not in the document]
xuannam230201: Good to know that! 

However, currently, I see that `FullKey` and `Stat.Key` in `detailed_metric` are not the same (because [current code](https://github.com/envoyproxy/ratelimit/blob/99d855108d0b97f3eec29f88dd788b34adbd4139/src/config/config_impl.go#L366-L368) doesn't set `FullKey` to `Stat.Key` in `detailed_metric`).

I set `FullKey` to `Stat.Key` in both cases (`detailed_metric` and `value_to_metric`), and also added unit tests for these cases. Is it okay?

xuannam230201:
@collin-lee, I updated PR to address your comments (and also added more unit tests). Could you please help to take a look at it? Thanks in advance!

xuannam230201:
Hi @collin-lee , I noticed that a GitHub Action failed after this PR was merged: https://github.com/envoyproxy/ratelimit/actions/runs/19781727464/job/56682854936
.
Could you please take a look and let me know if something went wrong (or it was just flaky?) I’m happy to help fix it if needed.

It also appears that the workflow did not push a new image with [new tag](https://hub.docker.com/layers/envoyproxy/ratelimit/master/images/sha256-533d62b2558990a754ea065f8ff2e6944f1a2f67106f6a18dd7309664eb6bf6b)
 to Docker Hub. It created an image with `master` tag.

xuannam230201:
In that GHA, build for master tag was [success](https://github.com/envoyproxy/ratelimit/actions/runs/19781727464/job/56682854936#step:5:421), but for new tag ([6b4f3896](https://github.com/envoyproxy/ratelimit/actions/runs/19781727464/job/56682854936#step:5:2344)) was failed. 

I checked the main.yaml GHA and see that they (master and new tags) are applied the same [command](https://github.com/envoyproxy/ratelimit/blob/6b4f38963267bc34bd68088060371fe0d0d8f67c/.github/workflows/main.yaml#L34-L35) to build and push Docker image. As a result, I think that some tests were flaky which caused this issue. Does it make sense to you to re-trigger this GHA to see if it's success @collin-lee .

The [failed test case](https://github.com/envoyproxy/ratelimit/actions/runs/19781727464/job/56682854936#step:5:34946) didn't relate to this change 
image

PR #1005, opened 2025-11-19
feat(redis): allow separate auth for sentinel vs redis

fefifofums:
- Added new optional environment variables `REDIS_SENTINEL_AUTH` and `REDIS_PERSECOND_SENTINEL_AUTH` to specify different connection options
 - Allow separate sentinel and redis connection options

### DIFF
--- a/src/redis/driver_impl.go
+++ b/src/redis/driver_impl.go
@@ -70,7 +70,7 @@ func checkError(err error) {
 	}
 }
 
-func NewClientImpl(scope stats.Scope, useTls bool, auth, redisSocketType, redisType, url string, poolSize int,
+func NewClientImpl(scope stats.Scope, useTls bool, auth, sAuth, redisSocketType, redisType, url string, poolSize int,
 	pipelineWindow time.Duration, pipelineLimit int, tlsConfig *tls.Config, healthCheckActiveConnection bool, srv server.Server,
 	timeout time.Duration,
 ) Client {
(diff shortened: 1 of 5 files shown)


src/redis/driver_impl.go
```diff
@@ -133,7 +156,14 @@ func NewClientImpl(scope stats.Scope, useTls bool, auth, redisSocketType, redisT
 		if len(urls) < 2 {
 			panic(RedisError("Expected master name and a list of urls for the sentinels, in the format: <redis master name>,<sentinel1>,...,<sentineln>"))
 		}
-		client, err = radix.NewSentinel(urls[0], urls[1:], radix.SentinelPoolFunc(poolFunc))
+
+		sentConnOpt := radix.SentinelConnFunc(sdf)
+		opt := radix.SentinelPoolFunc(poolFunc)
```
marwanad: Thank you for this 🫡 

Chiming in since we had a fun day with Sentinel and ratelimit today and I came across this change.

This didn't work to fix all our Sentinel woes. The missing piece here is that `poolFunc` also needs the custom connection `sdf` for sentinel.

`SentinelConnFunc` only applies to the sentinel connection itself and the pool is what's used for primary and secondary connections

marwanad:
@collin-lee @fefifofums any chance we can revive and bring this in? Sentinel is broken for us with ratelimit without the above.

collin-lee:
@marwanad @fefifofums - we got conflicts

PR #1017, opened 2025-12-05
[ISSUE] Fix: Wildcard Stats Key Behavior Changes

xuannam230201:
# [ISSUE]: Wildcard Stats Key Behavior Changes

## Problem

Previously, when `value_to_metric` logic was added (in my another [PR](https://github.com/envoyproxy/ratelimit/pull/996)), the wildcard stats key behavior was accidentally changed. Wildcard patterns without `value_to_metric` or `share_threshold` were generating stats keys without the wildcard value pattern (e.g., `test-domain.wild` instead of `test-domain.wild_foo*`). This broke backward compatibility for wildcard configurations.

Additionally, when share_threshold was implemented, stats keys for wildcard patterns with share_threshold were also not including the wildcard pattern with * (e.g., test-domain.files instead of test-domain.files_files/*), which was inconsistent with the expected behavior for wildcard stats.

**Note**: The lack of unit tests for wildcard stats keys made it difficult to detect this behavioral change. Adding comprehensive unit tests can help avoid such issues in the future.

## Solution

The code now preserves the original wildcard stats behavior while properly handling `share_threshold` and `value_to_metric` flags according to the following rules:

1. **No `value_to_metric` OR `detailed_metric`**: Stats key includes wildcard pattern with `*` (e.g., `test-domain.wild_foo*`)
2. **`value_to_metric` OR `detailed_metric`**: Stats key includes actual runtime value (e.g., `test-domain.wild_foo1`)
3. **`share_threshold: true`**: Stats key always includes wildcard pattern with `*`, regardless of other flags (e.g., `test-domain.wild_foo*`)

This change ensures that:
1. When a wildcard is matched and neither value_to_metric nor share_threshold is enabled, the stats key includes the wildcard pattern (e.g., foo*) instead of just the key name (original behavior).
2. When share_threshold is enabled, the stats key consistently includes the wildcard pattern with * (e.g., test-domain.files_files/*), maintaining consistency with wildcard stats behavior.

## Test Coverage

Added & Updated comprehensive unit tests to prevent future regressions.

### DIFF
--- a/src/config/config_impl.go
+++ b/src/config/config_impl.go
@@ -403,51 +403,37 @@ func (this *rateLimitConfigImpl) GetLimit(
 		// Build value_to_metric metrics path for this level
 		valueToMetricFullKey.WriteString(".")
 		if nextDescriptor != nil {
-			// Check if share_threshold is enabled for this entry
+			// Determine the value to use for this entry
+			var valueToUse string
 			hasShareThreshold := shareThresholdPatterns[i] != ""
+
 			if matchedWildcardKey != "" {
-				// If share_threshold: always use wildcard pattern with * (e.g., "foo*")
-				// Else if value_to_metric: use actual runtime value (e.g., "foo1")
-				// Else: use wildcard pattern with * (e.g., "foo*")
+				// Matched via wildcard
 				if hasShareThreshold {
 					// share_threshold: always use wildcard pattern with *
-					valueToMetricFullKey.WriteString(entry.Key)
-					valueToMetricFullKey.WriteString("_")
-					valueToMetricFullKey.WriteString(shareThresholdPatterns[i])
+					valueToUse = shareThresholdPatterns[i]
 				} else if nextDescriptor.valueToMetric {
-					valueToMetricFullKey.WriteString(entry.Key)
-					if entry.Value != "" {
-						valueToMetricFullKey.WriteString("_")
-						valueToMetricFullKey.WriteString(entry.Value)
-					}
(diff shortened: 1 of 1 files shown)


xuannam230201:
Hi @collin-lee , this issue was introduced in my [another PR](https://github.com/envoyproxy/ratelimit/pull/996) (this [line](https://github.com/envoyproxy/ratelimit/pull/996/files#diff-ed3f958d8794e4db695a58fd24d4e0410ffa4e7112c2d96dd4f3b7301e7aa8fcR387) and this [line](https://github.com/envoyproxy/ratelimit/pull/996/files#diff-ed3f958d8794e4db695a58fd24d4e0410ffa4e7112c2d96dd4f3b7301e7aa8fcR422)). As a result, I create this PR to address this issue (to maintain backward compatibility for metrics of wild card). Could you please help to take a look at this?

Thanks in advance!

src/config/config_impl.go
```diff
@@ -403,57 +403,38 @@ func (this *rateLimitConfigImpl) GetLimit(
 		// Build value_to_metric metrics path for this level
 		valueToMetricFullKey.WriteString(".")
 		if nextDescriptor != nil {
-			// Check if share_threshold is enabled for this entry
```
xuannam230201: Refactor the logic to include value in stats. It's not necessary to repeat:
```
valueToMetricFullKey.WriteString(entry.Key)
valueToMetricFullKey.WriteString("_")
valueToMetricFullKey.WriteString(...)
```
several times in this logic

PR #1027, opened 2025-12-10
add support for response dynamic metadata

zirain:
xref: https://github.com/envoyproxy/envoy/issues/41339

add a new environment flag `RESPONSE_DYNAMIC_METADATA`, which return DynamicMetadata in response if it's `true`.

### DIFF
--- a/src/service/ratelimit.go
+++ b/src/service/ratelimit.go
@@ -253,10 +253,11 @@ func (this *service) shouldRateLimitWorker(
 }
 
 func ratelimitToMetadata(req *pb.RateLimitRequest) *structpb.Struct {
-	dm, _ := structpb.NewStruct(nil)
-	dm.Fields = make(map[string]*structpb.Value)
+	fields := make(map[string]*structpb.Value)
+
 	// Domain
-	dm.Fields["domain"] = structpb.NewStringValue(req.Domain)
+	fields["domain"] = structpb.NewStringValue(req.Domain)
+
 	// Descriptors
 	descriptorsValues := make([]*structpb.Value, 0, len(req.Descriptors))
 	for _, descriptor := range req.Descriptors {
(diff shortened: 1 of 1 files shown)


src/service/ratelimit.go
```diff
@@ -239,10 +243,67 @@ func (this *service) shouldRateLimitWorker(
...
+	if this.responseDynamicMetadataEnabled {
+		response.DynamicMetadata = ratelimitToMetadata(request)
+	}
+
 	response.OverallCode = finalCode
 	return response
 }
 
+func ratelimitToMetadata(req *pb.RateLimitRequest) *structpb.Struct {
+	dm, _ := structpb.NewStruct(nil)
```
collin-lee: for NewStruct, I think you can look at getClientNode in xds_grpc_sotw_provider.go as an existing pattern in this Project to avoid silently ignored errors:

```
suggestion
func ratelimitToMetadata(req *pb.RateLimitRequest) *structpb.Struct {
    fields := make(map[string]*structpb.Value)
    
    // Domain
    fields["domain"] = structpb.NewStringValue(req.Domain)
    
    // Descriptors
    descriptorsValues := make([]*structpb.Value, 0, len(req.Descriptors))
    for _, descriptor := range req.Descriptors {
        s := descriptorToStruct(descriptor)
        if s == nil {
            continue
        }
        descriptorsValues = append(descriptorsValues, structpb.NewStructValue(s))
    }
    fields["descriptors"] = structpb.NewListValue(&structpb.ListValue{
        Values: descriptorsValues,
    })
    
    // HitsAddend
    if hitsAddend := req.GetHitsAddend(); hitsAddend != 0 {
        fields["hitsAddend"] = structpb.NewNumberValue(float64(hitsAddend))
    }
    
    return &structpb.Struct{Fields: fields}
}
```

PR #1045, opened 2026-01-10
implement quota mode for soft rate limit check

yuzisun:
Implement the quota mode soft limit feature for rate limiting service. 
- Added quota_mode: true field to descriptors (similar to shadow_mode) 
- Core Rate Limiting Logic: When a quota mode limit is exceeded, Request is allowed (returns 200 OK, not 429), Metadata is set indicating which quotas were exceeded. 
- Added quotaModeViolations: Array of descriptor indices that exceeded quota limits, Added quotaModeEnabled: Boolean indicating if any limits have quota mode enabled. 

Use Cases:
  1. 📊 Usage Tracking: Monitor customer usage without blocking requests                                                                                                                                                               
  3. 📈 Analytics: Collect violation patterns for capacity planning                                                                                                                                                                    
  4. 🚨 Alerting: Trigger alerts when soft limits are reached                                                                                                                                                                          
  5. 🎛️ Graceful Degradation: Allow services to fallback to another backend when one backend quota limit is reached
  
example test:
```bash
curl -X POST http://localhost:8080/json -H "Content-Type: application/json" -d '{"domain": "test_domain", "descriptors": 
[{"entries": [{"key": "quota_limit", "value": "test"}]}]}'
```
```json
{"overallCode":"OK", "statuses":[{"code":"OK", "currentLimit":{"requestsPerUnit":3, "unit":"MINUTE"}}], 
"dynamicMetadata":{"descriptors":[{"entries":["quota_limit=test"]}], "domain":"test_domain", "quotaModeEnabled":true, "quotaModeViolations":[0]}}
```

### DIFF
--- a/src/config/config_impl.go
+++ b/src/config/config_impl.go
@@ -456,6 +456,7 @@ func (this *rateLimitConfigImpl) GetLimit(
 					Limit:          originalLimit.Limit,
 					Unlimited:      originalLimit.Unlimited,
 					ShadowMode:     originalLimit.ShadowMode,
+					QuotaMode:      originalLimit.QuotaMode,
 					Name:           originalLimit.Name,
 					Replaces:       originalLimit.Replaces,
 					DetailedMetric: originalLimit.DetailedMetric,


yanavlasov:
Looks good to me. Thanks.

@collin-lee will wait for your review as well, since you were reviewing latest PRs. Note this change is to support Quota for Envoy AI Gateway https://github.com/envoyproxy/ai-gateway/issues/1571

The difference from the existing rate limit logic, is that in quota mode request is allowed if any of the descriptors have quota.

Quota mode would be enabled at the domain level. I.e. all descriptors under a specific domain will either be in quota or rate limit mode.

/wait-any

PR #1054, opened 2026-02-01
add quota mode xds config

yuzisun:
update quota mode for xds config update following up the previous PR for adding the quota mode.
https://github.com/envoyproxy/ratelimit/pull/1045


yuzisun:
@yanavlasov @collin-lee can you help review this ? It is needed to propagate the quota mode xDS config from ai gateway control plane to RLS.

PR #1059, opened 2026-02-05
Fix quota result when all limits were exceeded

yanavlasov:
Fix results in quota mode when all limits have been exceeded. In this case request should be denied.
Specifically rate limit service will deny the request under following conditions:

- If at least one descriptor in rate-limit mode had failed (this is the existing behavior and it did not change).
- Or all quote mode descriptors had failed (this is new behavior introduced in this PR).


yanavlasov:
@yuzisun for initial review

yanavlasov:
Need to rework the mixed mode. Will ping when it is ready for review.

agrawroh (approved):

Issue #728, opened 2024-10-09
Response time histograms using the prometheus sink are no longer in seconds


harpunius:
We migrated from statsD to the prometheus sink and use the following mapper snippet to monitor our infrastructure:

```yaml
    - match: "ratelimit_server.*.response_time"
      name: "ratelimit_service_response_time_seconds"
      timer_type: histogram
      labels:
        grpc_method: "$1"
```

These metrics used to be output in seconds, but are now output in ms.

As stated in the statsd-exporter [README](https://github.com/prometheus/statsd_exporter/blob/c18857b71b4afc2c304e4d34aa431a41234843ac/README.md?plain=1#L377-L380):

>Statsd timer data is transmitted in milliseconds, while Prometheus expects the unit to be seconds. The exporter converts all timer observations to seconds. Histogram and distribution events (`h` and `d` metric type) are not subject to unit conversion.

This used to happen when parsing observer events https://github.com/prometheus/statsd_exporter/blob/c18857b71b4afc2c304e4d34aa431a41234843ac/pkg/line/line.go#L82. In the new implementation, the histogram value is taken as-is: https://github.com/envoyproxy/ratelimit/blob/28b1629a21e885bdd2b527d6a1c1de8483dc47d4/src/stats/prom/prometheus_sink.go#L157.

This change (regression?) means that the default histogram buckets no longer make sense. I think we need to implement the same kind of unit switch. 

WDYT?

juliorenner:
this made me lose soooo much time trying to optimize rate limit service performance...

PR #1104, opened 2026-04-09
Fix Prometheus response time units

### LINKED ISSUE #1101: Prometheus histogram `ratelimit_service_response_time_seconds` is in fact observing milliseconds
In https://github.com/envoyproxy/ratelimit/pull/681, @zirain added Prometheus support to this searvice. The default mapper defines `ratelimit_service_response_time_seconds` based on `ratelimit_server.*.response_time`, which is in fact observing millisecond values:

https://github.com/envoyproxy/ratelimit/blob/05c08d0394737a24729e01519629420a4d86b865/src/metrics/metrics.go#L43

So, the name is wrong.

evevseev:
Fixes #1101

## Summary
- scale the built-in Prometheus `ratelimit_server.*.response_time` mapping from milliseconds to seconds so `ratelimit_service_response_time_seconds` matches its name and Prometheus unit conventions
- add `PROMETHEUS_RESPONSE_TIME_AS_MILLISECONDS=true` as a compatibility switch for users who rely on the legacy built-in behavior
- document the new compatibility flag and add tests for both the default and legacy behavior

## Breaking Change / Migration
The built-in Prometheus mapper now exports `ratelimit_service_response_time_seconds` in seconds instead of milliseconds.
Users who depend on the legacy built-in behavior can preserve it by setting:
```bash
PROMETHEUS_RESPONSE_TIME_AS_MILLISECONDS=true
```
This compatibility flag only affects the built-in default mapper. If PROMETHEUS_MAPPER_YAML is set, that custom mapping remains unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1105,6 +1105,7 @@ To enable Prometheus integration set:
 2. `PROMETHEUS_ADDR`: The port to listen on for Prometheus metrics. Defaults to `:9090`
 3. `PROMETHEUS_PATH`: The path to listen on for Prometheus metrics. Defaults to `/metrics`
 4. `PROMETHEUS_MAPPER_YAML`: The path to the YAML file that defines the mapping from statsd to prometheus metrics.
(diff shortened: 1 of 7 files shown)


agrawroh (approved):

harpunius:
Thanks! This also fixes https://github.com/envoyproxy/ratelimit/issues/728.

PR #1110, opened 2026-04-15
feat: add zipkin b3 header propagation

isker:
We unfortunately chose the wrong path down the tech tree many years ago and now we are stuck on zipkin header propagation throughout our service call stack. Configure the tracer here to parse and propagate zipkin's b3 headers in addition to the more standard otel/w3c ones.

This isn't gated behind configuration as I think the risk is small; this could only break someone's usage if they were already propagating both sets of headers to this service, and those headers conflicted. Given that Envoy is the typical client of this service, and I don't think you can even configure Envoy to do multiple types of propagation simultaneously, I think the risk is low. (There is no risk the other way as this service doesn't propagate headers to any place.)


agrawroh (approved):

isker:
Thanks!

PR #1131, opened 2026-05-04
fix: pin golang:1.26.2 to multi-arch index digest (arm64 exec format error on :master)

harrisonharris-di:
## Summary

Pin the `golang:1.26.2` base image in `Dockerfile`, `Dockerfile.integration`, and `examples/xds-sotw-config-server/Dockerfile` to the **multi-arch OCI image index** digest `sha256:b54cbf583d390341599d7bcbc062425c081105cc5ef6d170ced98ef9d047c716`, replacing the **single-arch (linux/amd64) image manifest** digest `sha256:7095ad02810845fa35d1fb090b8e57dd20dce4ca36b29b42951802350d2ec90e` introduced in #1124.

| Image | Digest | Type |
|---|---|---|
| `golang:1.26.1` (pre-#1124) | [`e2ddb153…`](https://hub.docker.com/layers/library/golang/1.26.1/images/sha256-e2ddb153f786ee6210bf8c40f7f35490b3ff7d38be70d1a0d358ba64225f6428) | OCI image **index** (multi-arch) |
| `golang:1.26.2` (current `master`, broken on arm64) | [`7095ad02…`](https://hub.docker.com/layers/library/golang/1.26.2/images/sha256-7095ad02810845fa35d1fb090b8e57dd20dce4ca36b29b42951802350d2ec90e) | OCI image **manifest** (linux/amd64 only) |
| `golang:1.26.2` (this PR) | [`b54cbf58…`](https://hub.docker.com/layers/library/golang/1.26.2/images/sha256-b54cbf583d390341599d7bcbc062425c081105cc5ef6d170ced98ef9d047c716) | OCI image **index** (multi-arch, includes the `7095ad02…` amd64 child) |

## Problem

Since #1124 merged on 2026-04-30, every `envoyproxy/ratelimit:master` image published to Docker Hub contains the linux/amd64 binary inside the linux/arm64/v8 layer. On arm64 hosts the container fails immediately at startup:

    exec /bin/ratelimit: exec format error

The arm64 layer in the published index is byte-identical to the amd64 layer at 9192208 bytes, vs the previous good build's separate 9151374 / 8398114 byte layers.

## Root cause

The previous 1.26.1 pin used the **multi-arch OCI image index** digest `sha256:e2ddb153f786...` (`mediaType: application/vnd.oci.image.index.v1+json`), which fans out per-platform when buildx is asked for arm64.

[...]

PR #1134, opened 2026-05-07
fix: pin golang:1.26.2 to multi-arch index digest

### LINKED ISSUE #1133: linux/arm64 latest (fe26676d) docker image not working
The latest envoyproxy/ratelimit docker image (tag `fe26676d` and `master`) does not work for the linux/arm64 architecture.
When running it on an arm64 machine, it just logs `exec /bin/ratelimit: exec format error`. Maybe the image was built only for amd64 and incorrectly tagged for both?

jukie:
Fixes #1133

PR #1124 pinned `FROM golang:1.26.2` by a per-platform (amd64) manifest digest instead of the OCI index digest. 

Under buildx + QEMU this caused every target platform to receive amd64 base layers, producing identical amd64 binaries inside the linux/arm64/v8 manifest and surfacing as `exec /bin/ratelimit: exec format error` on arm64 hosts.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.26.2@sha256:7095ad02810845fa35d1fb090b8e57dd20dce4ca36b29b42951802350d2ec90e AS build
+FROM golang:1.26.2@sha256:b54cbf583d390341599d7bcbc062425c081105cc5ef6d170ced98ef9d047c716 AS build
 WORKDIR /ratelimit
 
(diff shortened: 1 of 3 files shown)


jukie:
CC @collin-lee @bloomenergyguy who were involved with #1124

jukie:
Closing in favor of https://github.com/envoyproxy/ratelimit/pull/1131. Didn't see that before opening mine.

PR #1140, opened 2026-05-16
feat: implement is_negative_hits in rate limit descriptor

### LINKED ISSUE #41219: Enable per-user concurrency limits by supporting negative cost in the rate limiting protocol
*Title*: Enable per-user concurrency limits by supporting negative cost in the rate limiting protocol

*Description*:
Add a `hits_subtrahend` field to the rate limiting (RLS) protocol that acts like the `hits_addend` field, but decreases the rate limit counter instead of increasing it. Together with the `apply_on_stream_done` flag this would enable the emulation of per-user concurrency limits.

Goal: Ensure that each client can only have n requests open at the same time. This ensures fair resource allocation among clients and protects against "accidental DoS", and is simpler to implement and to reason about than cost-based limits.

Approach: Use a rate limit counter to track the number of requests the user has currently open by incrementing the counter at the beginning of the request (as usual with rate limiting) and then decrementing it again when the request is complete (using `apply_on_stream_done` with `hits_subtrahend`). This would limit the number of requests the client can perform in parallel, without limiting the total number per unit time.

Considerations:
* If envoy (or the node it runs on) terminates unexpectedly while a client request is ongoing, the counter will fail to be decremented, leading to a situation wherea client could potentially be "locked out". This is solved naturally by the fact that rate limit counters are reset (for fixed window) or decremented (for sliding window) at regular intervals. In the conztext of enforcing a concurrency limit, the time unit for which the limit is defined determines how often concurrency counts are "forgiven", allowing "stuck" clients to be unstuck.
* As a consequence of this mechanism it is possible for clients to exceed the provided limit at certain times up to a factor of two (assuming the connection timeout it smaller than the time window of the limit): if the client already has n requests in flight at the end of a time window, the counter will be reset and the client will be able to make another n requests even before the first set of requests is finished. This is no worse and the fact that, with fixed window semantics, a client can exceed the request count limit (or cost limit) by a factor of two if it makes requests just before and right after the end of a time window.

Example, with a limit of 2 per minute (per client):
1) start request 1, increment counter to 1
1) start request 2, increment counter to 2
1) start request 3, denied, too many requests
1) end request 2, decrement counter to 1
1) start request 4, increment counter to 2
1) ...envoy terminates unexpectedly, counter remains at 2
1) start request 5, denied, too many requests
1) start request 6, denied, too many requests
1) ...when the end of the 1 minute time window is reached...
1) counter is reset to 0
1) start request 7, increment counter to 1
1) end request 7, decrement counter to 0

rudrakhp:
## Summary

Add support for the `is_negative_hits` field on rate limit descriptors. When set, the
hits_addend value decrements the rate limit counter instead of incrementing it, effectively
refilling previously consumed tokens. The counter is floored at 0 (cannot go negative).

- Upgrade go-control-plane/envoy to v1.37.1 (adds IsNegativeHits proto field)
- Redis: uses an atomic Lua script (EVAL) to decrement and floor at 0 in a single operation
- Memcached: uses native Decrement which floors at 0 by default
- Negative hits bypass over-limit and near-limit checks (always return OK)
- New `total_negative_hits` stat tracks requested decrements per descriptor

Fixes https://github.com/envoyproxy/envoy/issues/41219

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/bradfitz/gomemcache v0.0.0-20230905024940-24af94b03874
 	github.com/coocood/freecache v1.2.4
 	github.com/envoyproxy/go-control-plane v0.14.0
-	github.com/envoyproxy/go-control-plane/envoy v1.37.0
+	github.com/envoyproxy/go-control-plane/envoy v1.37.1-0.20260409083702-98966259b99a
 	github.com/envoyproxy/go-control-plane/ratelimit v0.1.1-0.20260629194254-5581fade6089
 	github.com/go-kit/log v0.2.1
 	github.com/golang/mock v1.6.0
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@
     - [Health-check configurations](#health-check-configurations)
   - [GRPC server](#grpc-server)
 - [Request Fields](#request-fields)
+  - [Negative hits](#negative-hits)
 - [GRPC Client](#grpc-client)
   - [Commandline flags](#commandline-flags)
 - [Global ShadowMode](#global-shadowmode)
@@ -961,6 +962,21 @@ socket then set `GRPC_UDS`, e.g. `GRPC_UDS=/<dir>/ratelimit.sock` and leave
 For information on the fields of a Ratelimit gRPC request please read the information
 on the RateLimitRequest message type in the Ratelimit [proto file.](https://github.com/envoyproxy/envoy/blob/master/api/envoy/service/ratelimit/v3/rls.proto)
 
+## Negative hits
+
+Each descriptor entry may set the `is_negative_hits` field. When it is `true`, the descriptor's
+`hits_addend` is subtracted from the rate limit counter instead of being added to it, effectively
+refunding previously consumed capacity.
+
+Negative-hit behavior:
+
+- The counter is floored at `0` and can never go negative. Redis performs the decrement-and-floor
+  atomically via a Lua script; Memcached uses its native `Decrement`, which floors at `0` by default.
+- A negative-hit descriptor always returns `OK`. It bypasses the over-limit and near-limit checks and
+  does not trigger any over-limit side effects (over-limit stats or local-cache poisoning), even when
+  the counter is still above the limit after the decrement.
+- The requested decrement is counted in the `total_negative_hits` statistic (see [Statistics](#statistics-1)).
+
 # GRPC Client
 
 The [gRPC client](https://github.com/envoyproxy/ratelimit/blob/master/src/client_cmd/main.go) will interact with ratelimit server and tell you if the requests are over limit.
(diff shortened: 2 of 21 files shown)


arkodg:
cc @collin-lee @psbrar99

collin-lee (commented):
1. Redis Cluster Routing Bug (src/redis/fixed_cache_impl.go:41)

  pipelineAppendDecrement passes the Lua script text as the key parameter to PipeAppend. In Redis Cluster mode, PipelineAction.Key is used to determine which slot/node to route to. This means EVAL commands would be routed incorrectly — potentially causing
  MOVED/CROSSSLOT errors or silent misrouting when multiple negative-hit descriptors target different hash slots.

  2. Negative Hits Can Return OVER_LIMIT (src/redis/fixed_cache_impl.go:222)

  The PR claims "Negative hits bypass over-limit checks (always return OK)", but this is only true for the local-cache over-limit check. In the response loop, if the post-decrement counter is still above the threshold (e.g., counter at 15, limit 10,
  decrement 3 → result 12 > limit 10), GetResponseDescriptorStatus returns OVER_LIMIT. Same issue exists in memcached (src/memcached/cache_impl.go:146).

  3. Lua Script Returns new_val Instead of math.floor(new_val) (src/limiter/base_limiter.go:23)

  Minor inconsistency: the script stores math.floor(new_val) via SET but returns new_val directly. Practically identical for integer math but easy to fix for defensive correctness.

  Minor Issues

  - A decrement on a non-existent key creates a phantom key at value 0 with full TTL (functionally harmless but wastes memory)
  - No test coverage for negative hits with stopCacheKeyIncrementWhenOverlimit=true
  - No test for negative hits when the post-decrement value is still above the limit

  Suggestions

  1. Fix cluster routing by passing the actual Redis key (not the script) as the routing key for EVAL commands
  2. Force OK response for negative hits in the response handling loop (or document that OVER_LIMIT is possible)
  3. Add test cases for the above edge cases

rudrakhp:
@collin-lee thanks for the comments, couldn't get to this earlier, addressed them now.

rudrakhp:
@envoyproxy/ratelimit-maintainers gentle nudge, thanks :)

collin-lee (dismissed):

go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/bradfitz/gomemcache v0.0.0-20230905024940-24af94b03874
 	github.com/coocood/freecache v1.2.4
 	github.com/envoyproxy/go-control-plane v0.14.0
-	github.com/envoyproxy/go-control-plane/envoy v1.37.0
+	github.com/envoyproxy/go-control-plane/envoy v1.37.1-0.20260409083702-98966259b99a
```
collin-lee, line 11: Is this right?  Is there a non pseudo version?

collin-lee:
@rudrakhp - We should probably add README/docs updates for is_negative_hits and total_negative_hits.  Also see note about the go-control-plane version

go.mod
rudrakhp: There is no released version for `v1.37.1` or later, so we have to use this for now, till we get one released later.

collin-lee (approved):

PR #1144, opened 2026-05-25
feat: add retry in init phase instead of panic directly

### LINKED ISSUE #1143: Ratelimit Crashloop when starting
The Envoy rate limiting service should gracefully wait for the connection to its persistent storage(e.g. redis) to become available upon startup, rather than entering a crash loop.

1. We should retry instead of panic immediately
2. Marked pod as health after connection is ready

cc @collin-lee @psbrar99 WDYT?

zirain:
fixes: https://github.com/envoyproxy/ratelimit/issues/1143

Key Changes:
- Redis Connection Retry Mechanism: Replaces immediate panic on Redis connection failure with exponential backoff retry, validates connection with PING before proceeding.
- Context-Based Initialization: Adds context.Context throughout initialization chain for graceful cancellation, allows shutdown during startup retry phase.
- Health Check Updates: Redis health now starts as unhealthy (was healthy), marked healthy only after successful connection confirmed.


psbrar99:
LGTM, Thank you @zirain

psbrar99 (approved):

PR #1148, opened 2026-05-27
Add quota mode to rate limit descriptor proto

aabchoo:
Quota mode was added to the envoy/go-controlp-lane RateLimitDescriptor earlier this year: https://github.com/envoyproxy/go-control-plane/pull/1385/changes#diff-161ce5b2b49a210131bc9a8e7c14c1394a5a3a18f807e1eface73bbb401318f2

This PR updates the go-control-plane to include that change plus parses rl config message to include it.


aabchoo:
@agrawroh could you please take a look?

agrawroh (approved):

Issue #1132, opened 2026-05-06
High memory usage in metrics for RateLimitOverride


andreaaliu:
Hi, we're seeing very high memory usage on the metrics for rate limit requests sent with a rate limit override. Currently for rate limit override the value of entries is included in the stats key https://github.com/envoyproxy/ratelimit/blob/main/src/config/config_impl.go#L679-L691. This creates high cardinality for these requests. Is there an option to choose which entry values to include in the stats keys? Something like https://github.com/envoyproxy/ratelimit#including-descriptor-values-in-metrics which exists for those entries in the configuration.

andreaaliu:
https://github.com/envoyproxy/ratelimit/pull/1153 created a PR to fix this here, would appreciate a review, thanks!

limbov8:
Gentle bump on this.

PR #1153, opened 2026-06-03
Add optional rate limit key config for Envoy override stats keys

andreaaliu:
## Summary
- Add optional `DESCRIPTOR_KEY_CONFIG` YAML to control which descriptor entry values are included in stats keys for Envoy limit overrides.
- When unset, behavior is unchanged (all non-empty entry values are included).
- Load config once at provider startup; wire through file and xDS config providers.
## Motivation
`value_to_metric` / `detailed_metric` apply to config-matched limits, but override descriptors always include every entry value in the stats key, which causes high metric cardinality.  https://github.com/envoyproxy/ratelimit/issues/1132


andreaaliu:
Hi @envoyproxy/ratelimit-maintainers, could I please get a review? Thanks!

PR #1202, opened 2026-07-31
fix: prevent crash when client sends undefined rate limit unit

khoanguyenxuan:
## Description

When a client sends a gRPC `RateLimitRequest` with a descriptor limit override (`RateLimitOverride`) whose `unit` field is omitted or set to `UNKNOWN` (0), the server crashes with `panic: should not get here`.

The panic originates in `UnitToDivider()` (`src/utils/utilities.go`), which had no default case. The `UNKNOWN` unit flows from `GetLimit()` (`src/config/config_impl.go`), where the client-supplied override unit was trusted blindly and wrapped into a `RateLimit` without validation.

This is a denial-of-service vector: any caller of the rate limit service (e.g. Envoy, or a direct gRPC client) can crash the whole process with a single malformed request.

## Changes

- `src/config/config_impl.go`: `GetLimit()` now validates the override unit. If it is `UNKNOWN` (undefined), the invalid override is ignored with a warning and lookup falls back to the configured rate limit rules.
- `src/utils/utilities.go`: `UnitToDivider()` no longer panics on unknown units; it logs an error and returns a safe default (1 second) as defense in depth.
- `test/config/config_test.go`: regression test verifying an override with `UNKNOWN` unit is ignored and the configured limit is returned instead of panicking.

## Test Plan

- `go test ./src/...` passes.
- `go vet ./src/...` passes.
- `gofmt` clean.
- Manual reproduction confirmed the server previously crashed (`panic: should not get here`) and now continues serving when an undefined unit is sent.

### DIFF
--- a/src/config/config_impl.go
+++ b/src/config/config_impl.go
@@ -457,21 +457,25 @@ func (this *rateLimitConfigImpl) GetLimit(
 	}
 
 	if descriptor.GetLimit() != nil {
-		rateLimitKey := descriptorKey(domain, descriptor)
 		rateLimitOverrideUnit := pb.RateLimitResponse_RateLimit_Unit(descriptor.GetLimit().GetUnit())
-		// When limit override is provided by envoy config, we don't want to enable shadow_mode
-		rateLimit = NewRateLimit(
-			descriptor.GetLimit().GetRequestsPerUnit(),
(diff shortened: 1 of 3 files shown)


khoanguyenxuan:
Closing to create PR in fork first.

PR #1205, opened 2026-08-05
redis: log recovery when a connection succeeds after a prior dial error

OS-kiranmalsetty:
## Summary
- Adds an `atomic.Bool` to `poolStats` that tracks whether the most recent Redis dial attempt failed.
- Logs an info-level "redis connection re-established after previous error" message the first time a connection succeeds after prior dial errors, giving operators an explicit recovery signal instead of only seeing repeated error logs.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./src/redis/...`
- [x] `go test ./test/redis/...`

### DIFF
--- a/src/redis/driver_impl.go
+++ b/src/redis/driver_impl.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/jpillora/backoff"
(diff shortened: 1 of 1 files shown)


OS-kiranmalsetty:
@collin-lee, can you please review when you have a chance?

collin-lee (approved):

PR #1206, opened 2026-08-06
Add calendar-aligned MONTH rate limit, gated by USE_CALENDAR_MONTH_RATE_LIMIT

OS-kiranmalsetty:
## Summary
- A `unit: month` rate limit is currently computed as a fixed `60*60*24*30` second window counted from the Unix epoch, so it neither aligns with real calendar months nor accounts for months of different lengths.
- Adds a `USE_CALENDAR_MONTH_RATE_LIMIT` setting (default `"false"`) that, when enabled, buckets `MONTH` cache keys by UTC calendar month, sets their TTL/expiration to the actual time remaining until month end, and reports that same value as the reset duration (`DurationUntilReset` / reset header).
- Defaults to `false` so existing `MONTH` limits keep their current (legacy, fixed 30-day rolling window) reset behavior unless explicitly opted in — this avoids silently changing when existing users' MONTH limits reset.
- `src/utils/time.go`: new `expiryUntilMonthEnd`, `MonthExpirationSeconds`, `MonthStartUnix` helpers.
- `src/utils/utilities.go`: `ExpirationSeconds` and `CalculateReset` take a `useCalendarMonth` flag and dispatch to the calendar-aware helpers for `MONTH` only when it's true.
- `src/limiter/cache_key.go`, `src/limiter/base_limiter.go`, `src/redis/fixed_cache_impl.go`, `src/memcached/cache_impl.go`, `src/redis/cache_impl.go`, `src/service/ratelimit.go`: thread the flag from `settings.Settings.UseCalendarMonthRateLimit` down to cache-key bucketing, TTL, and reset reporting.
- `test/config/basic_config.yaml`: adds a `key8` MONTH descriptor for integration test coverage.
- README documents the new `USE_CALENDAR_MONTH_RATE_LIMIT` setting.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofumpt`/`goimports` clean on all changed files
- [x] `go test ./test/...` (all packages pass)
- [x] New unit tests cover: calendar-month cache-key bucketing (Jan 1/Jan 31 same bucket, Feb 1 different), leap-year correctness, and that the flag being `false` reproduces the legacy fixed-divider behavior exactly (including not consulting the time source at all for MONTH, matching pre-existing behavior)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@
   - [Two Redis Instances](#two-redis-instances)
   - [Health Checking for Redis Active Connection](#health-checking-for-redis-active-connection)
   - [Recovering from a failover (READONLY errors)](#recovering-from-a-failover-readonly-errors)
+  - [Calendar-aligned MONTH rate limits](#calendar-aligned-month-rate-limits)
 - [Memcache](#memcache)
 - [Custom headers](#custom-headers)
 - [Tracing](#tracing)
--- a/README.md
+++ b/README.md
@@ -1396,6 +1396,19 @@ configured address and reaches the current master. The failing command still ret
 to the caller; only the connection handling changes. Applies to both the main and the
 per-second Redis clients.
 
+## Calendar-aligned MONTH rate limits
+
+1. `USE_CALENDAR_MONTH_RATE_LIMIT` : (default is "false")
+
+By default, a `unit: month` rate limit uses a fixed 30-day window counted from the Unix epoch,
+which does not line up with real calendar months (it drifts, and treats every month as 30 days
+regardless of its actual length).
+
+Setting `USE_CALENDAR_MONTH_RATE_LIMIT` to `"true"` switches `MONTH` limits to a true calendar
+month window instead: the cache key bucket, TTL/expiration, and reported reset time all cover
+the 1st through the last day of the month (UTC). This is opt-in because it changes when
+existing `MONTH` limits reset and is therefore not enabled by default.
+
 # Memcache
 
 Experimental Memcache support has been added as an alternative to Redis in v1.5.
--- a/src/limiter/base_limiter.go
+++ b/src/limiter/base_limiter.go
@@ -22,6 +22,9 @@ type BaseRateLimiter struct {
 	localCache                 *freecache.Cache
 	nearLimitRatio             float32
 	StatsManager               stats.Manager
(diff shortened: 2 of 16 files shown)


OS-kiranmalsetty:
@collin-lee, can you please review when you have a chance?

OS-kiranmalsetty:
@envoyproxy/ratelimit-maintainers, can you please review?

OS-kiranmalsetty:
@collin-lee, can you please review when you have a chance?

collin-lee:
@OS-kiranmalsetty see https://github.com/memcached/memcached/blob/master/doc/protocol.txt and https://github.com/memcached/memcached/wiki/Programming#expiration 

In first article look for "Expiration times" section

In the second article:

"Any time higher than 30 days is interpreted as a unix timestamp date. If you want to expire an object on january 1st of next year, this is how you do that."

in src/memcached/cache_impl.go:170-173 see how `Expiration: int32(expirationSeconds)` is set

31*86400 = 2,678,400 which is greater than 2,592,000 so for (Jan, Mar, May, Jul, Aug, Oct, Dec) where there are 31 days Memcache will treat this as 1970-02-01 and it will be treated as already expired / immediately evicted.

OS-kiranmalsetty:
> @OS-kiranmalsetty see https://github.com/memcached/memcached/blob/master/doc/protocol.txt and https://github.com/memcached/memcached/wiki/Programming#expiration
> 
> In first article look for "Expiration times" section
> 
> In the second article:
> 
> "Any time higher than 30 days is interpreted as a unix timestamp date. If you want to expire an object on january 1st of next year, this is how you do that."
> 
> in src/memcached/cache_impl.go:170-173 see how `Expiration: int32(expirationSeconds)` is set
> 
> 31*86400 = 2,678,400 which is greater than 2,592,000 so for (Jan, Mar, May, Jul, Aug, Oct, Dec) where there are 31 days Memcache will treat this as 1970-02-01 and it will be treated as already expired / immediately evicted.

@collin-lee, updated the PR and also added additional test cases.

collin-lee (dismissed):

collin-lee:
@OS-kiranmalsetty - some build failures with the tests in cache_impl_test.go and base_limiter_test.go

memcached.NewRateLimitCacheImpl and limiter.NewBaseRateLimit expecting another boolean argument

collin-lee (approved):

Issue #32, opened 2018-01-15
Rolling window limits? 


eugenemiretsky:
Hi! 
Does the library support rolling windows limits?
For example, if my limit is 60 per hour, there are 2 ways to go about it
1) If the user hits the limit, she has to wait for 1 hour before the limit resets. Making even one call during the one hour time period will extend the limit by another hour. For example: if I made 60 calls in 1 hour, and then make one call in 1:59, I will be rate limited, and will have to wait for until 2:59 for the limit it reset )
2) The limits get incremented as the time window slides, to always count only the events in the last hour. For example: if I made 60 calls in 1 hour (5 of which were in the first minute), and 1:01 I am allowed to make 5 more calls. 

Based on reading the code, I think only 1 is supported, but just want to make sure

mattklein123:
Windowed limits are not currently supported.

andrascz:
Would you be interested in something like https://github.com/brandur/redis-cell?
This practically shifts the rate limit algorithm implementation from the service to Redis, but I think could be implemented with minimal effort.

mattklein123:
@andrascz it might be interesting to support something like that, though I think to be useful we would need to support those commands with whatever sharding/cluster system is being used. A single redis host is not going to scale in most deployments.

andrascz:
I am not 100% sure as I did not try it on cluster, but based on https://github.com/brandur/redis-cell/issues/30, either it is already working or needs minimal tweaks.

walbertus:
@mattklein123 If the `redis-cell` support sentinel and cluster setup, will `envoyproxy/ratelimit` use it? Or you are inclined toward internal code implementation?

mattklein123:
> @mattklein123 If the redis-cell support sentinel and cluster setup, will envoyproxy/ratelimit use it? 

Can you describe more what you are proposing?

walbertus:
There are two possible approaches to support rolling window limits:
1. Use [redis-cell](https://github.com/brandur/redis-cell) then create a new cache impl that uses redis-cell commands
2. Implement rolling window limit logic in new cache impl with basic Redis commands

The first approach would create a dependency on `redis-cell` module and shifting implementation to Redis.
This is a tradeoff between creating dependency vs maintaining implementation, which approach does the maintainer prefer?

mattklein123:
In general I think (2) is better, but I have no objection to having an option that requires redis-cell as long as it is opt-in.

exagil:
[...]

PR #1214, opened 2026-08-18
feat: add redis_decay backend and soft-breach response header

ysolomon-plat:
Adds BACKEND_TYPE=redis_decay: a continuously-decaying counter (GCRA family) executed as an atomic Redis EVAL. Unlike the fixed-window redis backend, cache keys carry no window timestamp, so budgets never reset at a boundary and the 2x boundary-burst is impossible by construction. Honors per-descriptor hits_addend. Fail-open on Redis errors. Cluster-safe: EVAL routes by cache key.

Also adds SOFT_BREACH_HEADER_RATIO: when set (0..1), admitted requests whose remaining budget falls inside the soft band get an x-ratelimit-soft-breach: 1 response header, so clients can degrade before hard 429s. Computed generically from CurrentLimit/LimitRemaining, so it works with any cache backend.

Covered by test/redis_decay (real EVAL execution on miniredis: boundary immunity, decay recovery, flood behavior, corrupt data, backward clock, fail-open, hits_addend) and test/service/soft_breach_test.go. All existing tests pass.

Load-tested on Istio 1.30 / Gateway API: 3,950 rps sustained on one replica, +2.3ms p50 overhead, exact admission under concurrency.

Refs envoyproxy/ratelimit#32 (rolling window request, open since 2018).

### DIFF
--- a/src/redis_decay/cache_impl.go
+++ b/src/redis_decay/cache_impl.go
@@ -1,0 +1,139 @@
+// Package redis_decay implements limiter.RateLimitCache with a continuously
+// decaying counter (GCRA family) executed atomically in Redis via EVAL.
+// PLA-8128 fork probe: ports the production algorithm from depop-routing
+// common/openresty/lua/introspection/ratelimiting.lua. Unlike the fixed-window
+// backend there is no window timestamp in the cache key, so there is no
+// boundary at which the budget resets.
+package redis_decay
+
+import (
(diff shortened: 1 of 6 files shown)


ysolomon-plat:
Superseded by #1215 (same change, rebuilt with DCO sign-off and cleaned-up comments).

PR #1230, opened 2026-09-01
invalidate local cache on is_negative_hits

oclaw:
**Problem**

is_negative_hits (https://github.com/envoyproxy/ratelimit/pull/1140) lets a descriptor decrement previously consumed quota, but refunds are invisible to the local over-limit cache: once a replica cached a key as over limit, a refund that freed quota in Redis kept being answered OVER_LIMIT from every replica's local cache until the window ended which degrades the feature on long windows

**Suggested design**

- Introduce feature flag ENABLE_NEGATIVE_HITS (default off). While off, a request containing a negative-hit descriptor is rejected with UNIMPLEMENTED
- Refuse to boot the one unfixable combination: negative hits + memcached + local cache (this PR does not add any support for the invalidations with memcached as a backend). Memcached refunds without a local cache enabled keep working.
- Add Redis pub-sub backed invalidation routine: at-most-once guarantee, single added connection per replica + local cache gen-based guard for eliminating races between invalidation and stale insert

P.S. Agent also added some missing tests for touched components, look fine

### DIFF
--- a/src/limiter/base_limiter.go
+++ b/src/limiter/base_limiter.go
@@ -4,7 +4,6 @@ import (
 	"math"
 	"math/rand"
 
-	"github.com/coocood/freecache"
 	pb "github.com/envoyproxy/go-control-plane/envoy/service/ratelimit/v3"
 	logger "github.com/sirupsen/logrus"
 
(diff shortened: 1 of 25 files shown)


src/limiter/base_limiter.go
```diff
@@ -14,24 +13,12 @@ import (
 	"github.com/envoyproxy/ratelimit/src/utils"
 )
 
-// DecrementScript atomically decrements a rate limit counter, floored at 0.
-// If the key does not exist there is nothing to refund, so it returns 0 without
-// creating a phantom key.
-const DecrementScript = `
```
oclaw, line 20: This is not really a common limiter part, moved it closer to Redis backend impl

src/redis/local_cache_invalidator.go
oclaw, line 1: Reused driver.go as much as possible but this routine has quite independent connection cycle so decided not to bother too much

src/redis/fixed_cache_impl.go
```diff
@@ -192,9 +228,12 @@ func (this *fixedRateLimitCacheImpl) DoLimit(
 			expirationSeconds += this.baseRateLimiter.JitterRand.Int63n(this.baseRateLimiter.ExpirationJitterMaxSeconds)
 		}
 
-		client, p := this.selectPipeline(cacheKey, &pipeline, &perSecondPipeline)
+		client, p, onPerSecondRedis := this.selectPipeline(cacheKey, &pipeline, &perSecondPipeline)
 		if hitsAddends[i].IsNegative {
-			pipelineAppendDecrement(client, p, cacheKey.Key, hitsAddends[i].Value, &results[i], expirationSeconds)
+			// The subscriber listens only on the main Redis.
+			publishInvalidation := this.publishInvalidations && !onPerSecondRedis
```
oclaw, line 234: dedicated per second redis assume high load and redis-backed refunds are barely effective here, not doing them (and the invalidator only listens main redis at the moment)

oclaw:
@collin-lee as a first-time contributor I do not have an opportunity to request for the review it looks like, so pinging you here :)

rudrakhp (commented):
Thanks for the bug fix. Can we keep it focused on fixing the cache invalidation part though, any reason why we need the `ENABLE_NEGATIVE_HITS` flag?

oclaw:
@rudrakhp The reason is that the bug is still present - I did not support the same invalidation for the memcached backend case as I am not sure how to deliver it properly. So without ENABLE_NEGATIVE_HITS and with local overlimit cache enabled memcached-backed workload will silently keep working in a broken way

